PR #75, opened 2021-09-17
Use Yarn workspaces

kherock:
First, apologies for the large diff!

I'm interested in making some contributions to this project and potentially some of the long-term upkeep. I ran into a couple problems setting up this repo for the first time:
* The checked-in package-lock.json files reference a private JFrog Artifactory instance
  * Additionally, the package-lock.json files are version `1`. NPM 7 released last October uses the new version `2` format.
* The pinned version of the typedoc markdown plugin was `^3.10`, but v3.11 doesn't support typedoc v0.21

Since this project has multiple `package.json`s I've gone ahead and upgraded the project to Yarn 3 with workspaces. I'm also open to using pnpm or npm workspaces if you have a preference @pamapa. I upgraded typedoc to v0.22 and the Docusaurus site runs for me now.


pamapa:
Hey, thanks for the interest to help improving this library. To streamline and minimize work it might be a good idea to first open an issue and discuss fundamental stuff there. On the other hand you probably only played around and tried different stuff :-), which is nice.

I just recently ported the old library to tsdx. tsdx (https://tsdx.io/) is is a zero-config CLI and i used tsdx already for several other projects. Works like a charm, supports treeshaking, ... . So i have no intention to change this now unless there is a good reason to do so.

Regarding npm vs yarn. I am aware of the debate. For this library i have choose to use npm. The CI work perfect all setup the last few week. When you develop you can even use locally yarn. Lets keep this. npm also supports workspaces...

In you merge request i see some eslint fixes e.g. in the website folder. Can you make a separate merge request with those changes? Would be nice. Maybe you can improve the current eslint setup incrementally to cover that as well?

Also a separate merge request for update typedoc to v0.22 would be nice.

Maybe its time to switch to npm v7 (supports workspaces). Would be nice if you can make a merge request for this.

Regarding the targeting es2019, might makes sense, i also thought about that. Please make a separate merge request for this, this might need some further discussions anyhow.

pamapa:
There is really many good this inside this monster request :-) Please also make a separate merge request for "fix: restore Angular sample project functionality", but please stick with npm (switch to v7 is good).

kherock:
Will break these up, thanks for looking.

As far as the use of TSDX goes, I think there are some major problems with its output format:

* Type definitions are generated for classes in modules that don't actually exist in the distribution.
image
  * because of this, Intellisense will suggest imports to classes that are actually undefined variables at runtime.
* The CJS entrypoint needlessly chooses between a minified production bulid or a development build
  * The CJS entrypoint is barely needed since this library doesn't run in Node
  * Users importing your code via a bundler are already equipped to minify library code
  * Otherwise, if they're not using a browser, loading the minified build offers little benefit other than to obscure stack traces
* TSDX seems to requires additional configuration to relax the ES5 output format, I didn't have the patience to figure out how to do it, but it seems like their Babel transforms are responsible for it.

Beyond those problems, the zero-config argument isn't that strong. ESBuild can emit in ~50ms, no config beyond tsconfig is needed.`@microsoft/api-extractor` works just by running `tsc` (which you already have configuration for) and then rolls up those definition files into one. It has extra features beyond that, but the configuration is quite minimal.

kherock:
I'll see how NPM workspaces run for now. My biggest complaint so far is that it isn't smart enough to determine when the current folder is part of a workspaces project, e.g. running `npm install` inside a folder will generate an extra package-lock file with a new `node_modules` tree despite those parts existing at the project root.

pamapa:
i just checked esbuild, looks quite good and it is very active maintained! @PascalLuginbuehl what do you think?
In the meantime please make small separate merge request with the other goodies.

pamapa:
commit "fix: restore Angular sample project functionality" without yarn would be nice on a separate merge request...

kherock:
There are a ton of npm audit issues caused by that app, I'm going to see if I can get it updated to Angular 12 when I have time. It's been awhile since I've touched Angular

pamapa:
i guess this merge request can be closed, as all features have been merged with different merge requests...

kherock:
@pamapa aside from using Yarn, that's true. I'd still like to point out that, at least on my end, the Docusaurus website works here but not in `main` with npm linking the dependencies.

tim-stasse:
@pamapa, I'd also like to advocate for the use of yarn 3.

The main benefit to using yarn (over npm) is its [Plug'n'Play installation strategy](https://yarnpkg.com/features/pnp). Which when used with its [offline cache](https://yarnpkg.com/features/offline-cache), enables the use of a [zero installs philosophy](https://yarnpkg.com/features/zero-installs). Considering the nature of this library, these features are a must have — particularly from a software supply chain & security perspective!

I've been using yarn 2 since it was first released & I'll admit (at first) it had some rough edges that needed ironing out (it was completely [re-architected](https://yarnpkg.com/advanced/architecture) from v1). However, those rough edges have since been ironed out with the release of yarn 3 & I would highly recommend it as the first port of call for any js/ts project.

I hope you would reconsider, I'm more than happy to provide an extensive list of benefits if you don't think the one's I've mentioned are enough already, or if you need me to elaborate on why said features are so important. I'd also be more than happy to contribute & help with this project if you did change it to yarn 3 (not that I \*wouldn't\* contribute if you didn't 😉).

PR #78, opened 2021-09-19
Update tsconfig files

kherock:
This should be the last PR before I send over the big refactor with the esbuild changes that should actually have tangible implications on the the library output :^)

This fixes up some tsconfig problems, namely
* unit tests only used loose type checking since they have their own tsconfig that didn't extend from the main project's
* `tsc` was unnecessarily performing some linting tasks that are already handled by ESLint (`noImplicitReturns`, `noFallthroughCasesInSwitch`, `noUnusedLocals`, `noUnusedParameters`)
  * `consistent-return` is actually slightly more strict than `noImplicitReturns` in that doesn't find it acceptable to inconsistently return a `Promise<void>` in a void async function.
* I've also enabled `importsNotUsedAsValues` and fixed all instances of the rule. It's primary useful in the following scenario:
  ```typescript
  // this import needs to be stripped in the output as it doesn't import runtime code
  import { OtherClass } from './foo';
  
  export class MyClass {
    // this is difficult to tree-shake - the bundler must evaluate that `OtherClass` is unused
    // and then assert that `import './foo'` is pure
    foo: OtherClass;
  }
  ```

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -9,6 +9,7 @@ parserOptions:
 extends:
   - eslint:recommended
 rules:
+  consistent-return: error
   indent:
     - error
     - 4
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,5 +1,6 @@
 plugins:
   - "@typescript-eslint"
+  - testing-library
 env:
   browser: true
   node: true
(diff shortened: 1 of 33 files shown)


pamapa (approved):

pamapa:
Great work, thanks for contributing!

PR #89, opened 2021-09-22
fix: restore Angular sample project functionality

kherock:
This actually ended up being a pretty simple upgrade. I upgraded all Angular dependencies to version 12 and removed the old TSLint/codelyzer tooling in that part of the project. No code changes were really needed beyond some updates to the callback endpoint (oidc-client needs a hint for `response_mode: 'query'`).


kherock:
I'm really unsure why adding the angular workspace's dependencies is causing tsdx's `ts-jest` instance to fail with type resolution. I've managed to reproduce it locally once, but it went away after using a clean workspace.

pamapa:
pipeline must succeed. Lets merge first #90 (refactor build process to use esbuild and API Extractor) and see if that helps :-)

pamapa:
nice, thanks for contributing

pamapa (approved):

PR #74, opened 2021-09-17
feature PopupWindow postmessage

rvdkooy:
This PR uses postmessage logic to send data back to the main window

todo:
 - [x] use postmessage
 - [x] add tests

PR #90, opened 2021-09-23
refactor build process to use esbuild and API Extractor

kherock:
Summary of changes:
* `tsdx build` is replaced by build scripts for various targets using esbuild
  | entrypoint | description |
  | --- | --- |
  | `oidc-client-ts.cjs` | CommonJS target, useful for Node-based environments such as unit tests using Jest |
  | `oidc-client-ts.mjs` | ESM bundle - bundlers will prefer this export, and browsers that support  ESM can<br> import this directly from a CDN
  | `oidc-client-ts.js` | IIFE bundle that should be acceptable for most browser environments. The API is<br> exported under a global `oidc` namespace |
  | `oidc-client-ts.min.js` | minified version of the above |
* all of the above are compatible with environments implementing ES2019
  * as it is becoming the norm, IE compatibility is still feasible via bundlers that transpile and polyfill missing web features.
* [@microsoft/api-extractor](https://www.npmjs.com/package/@microsoft/api-extractor) is used for rolling up a `.d.ts` bundle
  * additionally, incremental TS compilation is now enabled for drastically faster builds. CI will need to cache `*.tsbuildinfo` files to take advantage of this.
* disabled type-checking in the `ts-jest` process in favor of plain incremental compilation with `tsc`
  * since the build process doesn't use `tsc` (and therefore requires `"isolatedModules": true"` in tsconfig), we don't need `tsc` for unit tests either.


pamapa:
nice work, regarding `etc/oidc-client-ts.api.md`, our first plan was to generate the API doc using typedoc , the API Extractor looks really promising and easy to use. The docusaurus part of the library is not finished yet, but i guess the correct place for the oidc-client-ts.api.md file should be inside the ./docs folder. The idea is to move API related stuff from ./docs/intro.md to the source code and only add add link to the ./docs/oidc-client-ts.api.md file.

api-extractor.json
```diff
@@ -0,0 +1,364 @@
...
+         * The API report file is normally tracked by Git.  Changes to it can be used to trigger a branch policy,
+         * e.g. for an API review.
+         *
+         * The path is resolved relative to the folder of the config file that contains the setting; to change this,
+         * prepend a folder token such as "<projectFolder>".
+         *
+         * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
+         * DEFAULT VALUE: "<projectFolder>/etc/"
+         */
+        // "reportFolder": "<projectFolder>/etc/",
```
pamapa, line 138: "reportFolder": "<projectFolder>/docs/"
->
./docs/oidc-client-ts.api.md

see other comment for more context...

.husky/pre-commit
```diff
@@ -2,3 +2,8 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+
+# update the API report
+mkdir -p ./etc
```
pamapa, line 7: needs adaption for using ./docs/oidc-client-ts.api.md

api-extractor.json
kherock: Sounds good, and makes sense.

Also, is the Docusaurus site building for you since the merge of all of those dependabot PRs? Both `docusaurus-plugin-typedoc` and `typedoc-plugin-markdown` needed updates to support Typedoc 0.22. The webpack build inconsistently reports between being unable to find `api/index.md` or that the sidebar can't render due to `api/index` being an unknown document name.

etc/oidc-client-ts.api.md
```diff
@@ -0,0 +1,450 @@
...
+```ts
+
+// @public (undocumented)
+export class AccessTokenEvents {
+    constructor({ expiringNotificationTimeInSeconds }: {
+        expiringNotificationTimeInSeconds: number;
+    });
+    // (undocumented)
+    addAccessTokenExpired(cb: AccessTokenCallback): void;
+    // Warning: (ae-forgotten-export) The symbol "AccessTokenCallback" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
pamapa: i guess yes, this can be done in a follow up merge request

api-extractor.json
pamapa: lets fix that i a separate merge request

.husky/pre-commit
```diff
@@ -2,3 +2,8 @@
...
+npm run build-types -- -l
+git add ./etc/oidc-client-ts.api.md
```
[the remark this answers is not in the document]
pamapa: I have the feeling that `git add ./docs/oidc-client-ts.api.md` will lead to merge conflicts or parts in a merge request we do not want to merge from "yet unknown" developers, then that developer needs to do a rebase, which not everybody can do, .... It might be better when we do this manually (e.g. when the CI fails) and particularly before we do a release (to refresh docs).

I am not sure if we want to have husky at all. First i was worried it wont work under Windows, but it looks like it does. But it duplicates stuff which the CI is already handling for us. Thus it makes stuff more complicated. husky might be good for private repos where all involved developers know about it, but this is an open repo and we will have merge request from outside...

Lets remove husky for now unless there is a real use case! Lets keep stuff simple and stupid.

tsconfig.json
```diff
@@ -2,29 +2,29 @@
   // see https://www.typescriptlang.org/tsconfig to better understand tsconfigs
   "include": ["src", "types"],
   "compilerOptions": {
-    "target": "ES5",
-    "module": "esnext",
-    "lib": ["dom", "esnext"],
-    "importHelpers": true,
+    "target": "ES2019",
```
pamapa, line 5: conservatively we could use here ES2018, according to TypeScript docu:
- ES2018 | Additional APIs available in ES2018 - **async iterables, promise.finally**, Intl.PluralRules, regexp.groups, 
- ES2019 | Additional APIs available in ES2019 - array.flat, array.flatMap, Object.fromEntries, string.trimStart, string.trimEnd, etc.

-> ES2019 does not bring much more for us against E2018... On the other hand its a new library and we have the year 2021...

.husky/pre-commit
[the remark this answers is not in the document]
kherock: Sure, but I at least think lint-staged is still useful at any scale, it's simply better dev ergonomics. CI isn't fixing and committing code that was edited, only your local machine can do that. For example #74 had some back and forth that could have been avoided. The CI step is definitely still needed, though, because merging and rebasing can give rise to lint errors that don't run through the pre-commit hook.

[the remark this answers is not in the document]
pamapa: ok, but lets keep it only for lint-staged...

tsconfig.json
kherock: I've been using `.flatMap` and `.fromEntries` a lot more frequently lately, but I understand the reasoning. API polyfills are very easy for end-users to add. ES2019 support came with Safari 12.1 (March 2019).

pamapa: ok, lets try with ES2019

src/Version.ts
```diff
@@ -1,2 +1,7 @@
-const Version = "2.0.0-alpha2";
-export { Version };
+// @ts-expect-error avoid enabling resolveJsonModule to keep build process simple
+import { version } from "../package.json";
```
pamapa, line 2: nice

.husky/pre-commit
```diff
@@ -2,3 +2,7 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+
+# update the API report
+npm run build-types -- -l
```
pamapa, line 7: lets from those

kherock:
Ok I removed the commit hook for API extractor. I've intentionally left the building failing for now just to show what CI looks like when the public API changes.

pamapa (approved):

pamapa:
really nice work, thanks for the contribution!

PR #96, opened 2021-09-27
Support passing window params to UserManager functions

### LINKED ISSUE #95: Restore a subset of the args removed from the sign in methods
With oidc-client 1.x, I used a few parameters in the `signinRedirect`, `signinPopup`, and `signinSilent` methods which were removed in #20:

* `extraQueryParams` - I need to set this dynamically to tweak the auth server's login page. For example, Keycloak has [`kc_idp_hint`](https://www.keycloak.org/docs/latest/server_admin/#_client_suggested_idp) that allows me to create multiple buttons for launching different identity provider login flows. I also might need to be able to suggest a username that's only obtained after the UserManager is created
* `popupWindowFeatures`
  * I want to be able to center the popup window in front of the open window rather than the hardcoded offset that's set when constructing the UserManager. These offsets can only be accurately determined the moment the popup is opened.
  * I need to be able to customize the size of the window based on various login pages that can show based on the parameter I mentioned above

kherock:
Resolves #95.

This restores the ability to customize the window behavior of the Popup, IFrame, and Redirect navigators. The existing `NavigatorParams` interface was taking on too many parameters that weren't related to the actual `.navigate()` step during sigin/signout. I've split the following settings out of that interface and made them customizable via params passed to the `.signin*()`/`.signout*()` methods:

* `popupWindowFeatures`
* `popupWindowTarget`
* `silentRequestTimeoutInSeconds`
* `redirectMethod`

This PR has some clean up commits, I recommend reviewing 678ef18 in isolation for a clear view of the refactoring for the above settings.

These changes contain one potentially breaking change where the popup's redirect method is no longer customizable and uses the `'replace'` method instead of `'assign'`. This was chosen because in a popup, navigating backwards would load an un-actionable blank page.

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -27,7 +27,7 @@ export class AccessTokenEvents {
 
 // @public (undocumented)
 export class CheckSessionIFrame {
-    constructor(callback: () => Promise<void> | void, client_id: string, url: string, intervalInSeconds: number, stopOnError: boolean);
+    constructor(_callback: () => Promise<void>, _client_id: string, url: string, _intervalInSeconds: number, _stopOnError: boolean);
     // (undocumented)
     load(): Promise<void>;
     // (undocumented)
@@ -208,19 +208,19 @@ export interface OidcClientSettings {
 export class SessionMonitor {
     constructor(userManager: UserManager);
     // (undocumented)
-    protected _callback(): Promise<void>;
+    protected _callback: () => Promise<void>;
     // (undocumented)
     protected _init(): Promise<void>;
     // (undocumented)
-    protected _start(user: User | {
+    protected _start: (user: User | {
         session_state: string;
         profile: {
             sub: string;
             sid: string;
         } | null;
-    }): Promise<void>;
+    }) => Promise<void>;
     // (undocumented)
-    protected _stop(): void;
+    protected _stop: () => void;
 }
 
 // @public (undocumented)
@@ -310,7 +310,7 @@ export class UserManager {
     // (undocumented)
     protected readonly _popupNavigator: PopupNavigator;
     // (undocumented)
-    querySessionStatus(): Promise<SessionStatus | null>;
+    querySessionStatus({ silentRequestTimeoutInSeconds, ...args }?: IFrameWindowParams & ExtraSigninRequestArgs): Promise<SessionStatus | null>;
     // Warning: (ae-forgotten-export) The symbol "RedirectNavigator" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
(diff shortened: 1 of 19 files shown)


pamapa:
the oidc-client-ts.api.md must be generated + copied + added :-)

src/UserManager.ts
```diff
@@ -113,24 +112,23 @@ export class UserManager {
...
+        const handle = await this._popupNavigator.prepare({
             popupWindowFeatures: this.settings.popupWindowFeatures,
             popupWindowTarget: this.settings.popupWindowTarget,
-            redirectMethod: this.settings.redirectMethod
+            ...params
         });
+        const user = await this._signin({
+            request_type: "si:p",
+            redirect_uri: url,
+            display: "popup"
```
pamapa, line 130: hmm original args/params was passed into _signin and not navigatorParams...
See  https://github.com/IdentityModel/oidc-client-js/blob/d700897e0a9e5eac060b12a687d55606535ca5e9/src/UserManager.js#L127

src/UserManager.ts
```diff
@@ -152,30 +150,52 @@ export class UserManager {
...
+
         if (user && this.settings.validateSubOnSilentRenew) {
             Log.debug("UserManager.signinSilent, subject prior to silent renew: ", user.profile.sub);
             args.current_sub = user.profile.sub;
         }
 
-        return this._signinSilentIframe(args);
+        const handle = await this._iframeNavigator.prepare({
+            silentRequestTimeoutInSeconds: this.settings.silentRequestTimeoutInSeconds,
+            ...params,
```
pamapa, line 180: same here

src/UserManager.ts
```diff
@@ -152,30 +150,52 @@ export class UserManager {
         }
     }
 
-    public async signinSilent(): Promise<User | null> {
+    public async signinSilent(params?: IFrameWindowParams): Promise<User | null> {
```
pamapa, line 153: can we rename that argument and  `params` to `args` including the others?
args -> belongs to create request
navigatorParams -> belongs to navigation

Also i guess the type is wrong here, it must be SigninArgs

src/UserManager.ts
```diff
@@ -287,26 +281,25 @@ export class UserManager {
         }
     }
 
-    public async querySessionStatus(): Promise<SessionStatus | null> {
+    public async querySessionStatus(params?: IFrameWindowParams): Promise<SessionStatus | null> {
```
pamapa, line 284: type must be SigninArgs

src/UserManager.ts
```diff
@@ -287,26 +281,25 @@ export class UserManager {
...
         const url = this.settings.silent_redirect_uri || this.settings.redirect_uri;
         if (!url) {
             Log.error("UserManager.querySessionStatus: No silent_redirect_uri configured");
             throw new Error("No silent_redirect_uri configured");
         }
 
-        const args = {
+        const handle = await this._iframeNavigator.prepare({
+            silentRequestTimeoutInSeconds: this.settings.silentRequestTimeoutInSeconds,
+            ...params
```
pamapa, line 293: original it was passed in to create request and not navigator...

src/UserManager.ts
```diff
@@ -386,14 +378,14 @@ export class UserManager {
...
+    protected async _signinCallback(url: string | undefined, navigator: IFrameNavigator | PopupNavigator): Promise<void> {
         Log.debug("UserManager._signinCallback");
         const useQuery = this.settings.response_mode === "query" || (!this.settings.response_mode && SigninRequest.isCode(this.settings.response_type));
         const delimiter = useQuery ? "?" : "#";
-        return navigator.callback(url, false, delimiter);
+        await navigator.callback(url, false, delimiter);
     }
 
-    public async signoutRedirect(): Promise<void> {
+    public async signoutRedirect(params?: RedirectParams): Promise<void> {
```
pamapa, line 388: same here

src/UserManager.ts
```diff
@@ -411,7 +407,7 @@ export class UserManager {
         return response;
     }
 
-    public async signoutPopup(): Promise<void> {
+    public async signoutPopup(params?: PopupWindowParams): Promise<void> {
```
pamapa, line 410: same here

src/UserManager.ts
kherock: SigninArgs doesn't include the window params that I'm passing here, I haven't dealt with those settings yet. As of this PR, it's not possible to customize those with function arguments. The name `params` was chosen for consistency with the naming with the `NavigatorParams` interface that I've split into `PopupWindowParams`, `RedirectParams`, and `IFrameWindowParams`.

src/UserManager.ts
kherock: Could you clarify a bit? In the linked line, I've combined the 2nd and 3rd args to create the `handle` instance that I pass to `_signin`. The object passed to `_signin` is the same object that was previously declared on its own a few lines before.

pamapa: in order to restore original functionality (https://github.com/IdentityModel/oidc-client-js/blob/dev/src/UserManager.js#L127), how it was:
public signinRedirect(**args**) -> this._signinStart(**args**,... -> this.createSigninRequest(**args**)

https://github.com/pamapa/oidc-client-ts/blob/456addb3ea4e247a5e43f1cf813dd7da04fc3bf3/src/UserManager.ts#L93:
You are now passing **by default** arguments to the navigator and _only_ extraQueryParams, extraTokenParams. To restore functionality it might be better to have the sigin arguments the default and handle 

Your code:
```
signinRedirect({
        extraQueryParams,
        extraTokenParams,
        **...navigatorParams**
    }: ExtraSigninRequestArgs & RedirectParams = {}
```
-> would restore not mean? With this approach its possible to inject more as sigin request args... even if we do not explicitly declare the types...
```
signinRedirect({
    redirectMethod?: "replace" | "assign",
    **...args**
}: ExtraSigninRequestArgs & RedirectParams = {}
```

-> ~~Maybe its better to not mix up sigin arguments with navigation parameters on the API at all. As such we could avoid what the _default_ is. This makes the API more difficult to use.~~
```
**not** signinRedirect(args?: ExtraSigninRequestArgs = {}, navigatorParams?: RedirectParams = {})
```

kherock: Ok - so as far as passing extra navigator params "by default", I don't see it as an issue because I do explicitly type the extent of what can be passed to those methods, but I can reverse them you really think it improves readability.

On being able to inject more signin request args, I really do think that the ones chosen (`extraQueryParams` and `extraTokenParams`) are all that are necessary, adding more would overcomplicate the function signature. The code I have now causes a type error when passing additional properties.
image


I considered splitting up the objects like you say here
> -> Maybe its better to not mix up sigin arguments with navigation parameters on the API at all. As such we could avoid what the default is. This makes the API more difficult to use.
> <pre><code>**not** signinRedirect(args?: ExtraSigninRequestArgs = {}, navigatorParams?: RedirectParams = {})</code></pre>

and I still like the idea, but it's definitely easier to work with a function that only takes 1 argument vs 2.

pamapa: please reverse them, lets keep args the default (even if we have only extraQueryParams and extraTokenParams) and navigiation params shall be the special extra handled ones

kherock: I added a commit testing what two parameters would look like, and I don't hate it. I still feel like customizing the request args with `extraQueryParams` is the exceptional case. Your suggestion is reasonable though, I'll reverse the grouping and am happy to leave it at that

docs/oidc-client-ts.api.md
```diff
@@ -310,7 +310,7 @@ export class UserManager {
     // (undocumented)
     protected readonly _popupNavigator: PopupNavigator;
     // (undocumented)
-    querySessionStatus(): Promise<SessionStatus | null>;
+    querySessionStatus(navigatorParams?: IFrameWindowParams, args?: ExtraSigninRequestArgs): Promise<SessionStatus | null>;
```
pamapa, line 313: to restore previous behavior and to have it the same way as the API functions signinPopup, signinRedirect: lets combine here the args as well like:
`{ silentRequestTimeoutInSeconds, ...args }?: IFrameWindowParams & ExtraSigninRequestArgs`

pamapa (approved):

PR #100, opened 2021-09-27
Use standard-version for releases

kherock:
These changes update the existing publish workflow to instead run when tags are pushed after running the standard-version script. A sample run of these workflows is viewable on my fork [here](https://github.com/kherock/oidc-client/actions/runs/1280156682).

The scripts I've written include support for prerelease tags. If there's no intent to make prereleases after the initial 2.0 release, I can probably simplify them a bit.


pamapa:
Maybe there will be pre-releases even after 2.0.0. Maybe we make a 2.1.0 and need to pre-test stuff. I have seen other projects doing that, why not? Depends on the future changes...

kherock:
@pamapa is there anything else I should include here? I was thinking that a workflow trigger for creating a release could be useful.

pamapa:
Yes something i really like in lerna (https://github.com/lerna/lerna) is semi-automatic releasing. you can interactive choose the version, it adapts files and makes a commit and pushes. Just for the record lerna is for multi-project, which this project is not :-)

Currently i coupled the publish with the github publish, which is also quiet nice, as we can make a tag, control the tag and then manually create a npm registry release by creating a github release.

kherock:
Yeah Lerna is definitely intended for multi project setups. I think this project can operate just fine with automatic version tagging implemented here, so I wouldn't recommend it. Maybe a pull_request action enforcing commit message formats with commitlint would be good?

pamapa:
lets keep things simple, other developer shall have it easy to provide features/fixes...

pamapa:
i guess this MR can be closed, the CI workflows have been renamed and updated. For the release making i am researching a helper library see #272

PR #105, opened 2021-09-30
fix lint warnings and exports

pamapa:
Additional to the lint fixes, i tweaked the API and we need to export more.

esbuild is more conservative than tsdx regarding exporting, which is good. I am for now only exporting what is needed by the context library. We might need to export more, we can do this later when requested.

In order to pass-through the APT arguments for signinRequest, ... i have introduced separate types, this makes it much more easy.

All seen when trying to use this library here: https://github.com/pamapa/react-oidc-context/pull/78

### DIFF
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -49,7 +49,7 @@ export class PopupWindow implements IWindow {
 
             this._id = params.id;
             if (this._id) {
-                (window as any)["popupCallback_" + params.id!] = this._callback;
+                (window as any)["popupCallback_" + this._id] = this._callback;
             }
 
             this._popup.focus();
@@ -113,7 +113,10 @@ export class PopupWindow implements IWindow {
 
         window.removeEventListener("message", this._messageReceived);
 
-        delete (window as any)["popupCallback_" + this._id!];
+        if (this._id) {
+            delete (window as any)["popupCallback_" + this._id];
+        }
+        this._id = undefined;
 
         if (this._popup && !keepOpen) {
             this._popup.close();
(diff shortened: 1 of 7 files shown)


src/navigators/RedirectNavigator.ts
```diff
@@ -26,7 +26,7 @@ export class RedirectNavigator implements INavigator, IWindow {
             throw new Error("No url provided");
         }
 
-        window.location[this._redirectMethod!](params.url);
+        window.location[this._redirectMethod || "assign"](params.url);
```
kherock, line 29: This is actually a valid usage of `!`. `_redirectMethod` must be defined by this point since the caller should have used `INavigate#prepare` to obtain an `IWindow` instance with a `#navigate` implementation. I think a better way to get rid of this warning is to refactor the `IWindow` implementation into a `RedirectWindow` that matches the other navigators.

src/navigators/PopupWindow.ts
```diff
@@ -49,7 +49,7 @@ export class PopupWindow implements IWindow {
 
             this._id = params.id;
             if (this._id) {
-                (window as any)["popupCallback_" + params.id!] = this._callback;
+                (window as any)["popupCallback_" + this._id] = this._callback;
```
kherock, line 52: 👍

kherock (commented):
Is there a sounder way to keep track of these callback functions aside

src/navigators/RedirectNavigator.ts
pamapa: The compiler complains because its not initialized in the constructor, which then leads to a null assertion. In theory the prepare method should be called before. Until we refactor, this fix should not harm as at that point this._redirectMethod should have the proper value...

PascalLuginbuehl (approved):

PR #149, opened 2021-10-19
Reduce usage of any across some files

TomCaserta:
Just getting acquainted with the repo and thought I'd assist with #4 a little, heres some quick low hanging fruit with regards to using any. Feel free to close if not fitting your guidelines of course. 

The events have a rather odd generic syntax to them as the raise method supposedly accepts an arbitrary amount of parameters although from what I can tell this is only actually utilised in the tests?


pamapa:
nice work, thanks for your contribution!

PR #150, opened 2021-10-19
Remove ts-ignore from random implementation

TomCaserta:
Another small one removing ts-ignore. 

I removed the "magic numbers" that create the uuid template string, it didn't really reduce character count previously by much considering it was repeated in both uuid implementations.

That being said could probably refactor this further as it doesn't make sense that it creates a UUID containing `-` then effectively removes them in the only exported function of `random`.


TomCaserta:
Changes made and should be building fine now. Probably will want to squash merge via the github merge options, alternatively I can rebase.

pamapa:
changes look good, yes please rebase, iff i squash you are no longer the author...

PR #162, opened 2021-10-26
improve response_mode handling

pamapa:
- [x] i will drop most part of response_mode as we only need query now that implicit has gone

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -67,8 +67,6 @@ export interface CreateSigninRequestArgs {
     // (undocumented)
     resource?: string;
     // (undocumented)
-    response_mode?: string;
-    // (undocumented)
     response_type?: string;
     // (undocumented)
     scope?: string;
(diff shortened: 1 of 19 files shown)


docs/oidc-client-ts.api.md
```diff
@@ -65,8 +65,6 @@ export interface CreateSigninRequestArgs {
     // (undocumented)
     resource?: string;
     // (undocumented)
-    response_mode?: string;
```
longsleep, line 68: `response_mode` should not be removed. I do not agree with the statement that it is no longer needed since implicit was removed.

Using response mode `fragment` with code flow is also beneficial as it prevents sending things like "code" to the server which actually serves the application.

longsleep: I would not be able to use `oidc-client-ts` if it should loose support for `fragment` response mode.

longsleep (changes requested):

docs/oidc-client-ts.api.md
pamapa: ok, lets keep this feature

PR #176, opened 2021-11-02
default response mode is explicit query

pamapa:
- response_mode is kept
- by made easier, by setting by default query
- fragment response_mode is still possible


longsleep (approved):
LGTM

PR #192, opened 2021-11-03
move text from index.md into source code

pamapa:


src/OidcClientSettings.ts
```diff
@@ -42,14 +42,24 @@ export interface OidcClientSettings {
     redirect_uri: string;
     /** The OIDC/OAuth2 post-logout redirect URI */
     post_logout_redirect_uri?: string;
+
+    /** Indicates when sending client secret if sent as a post param or in `Authorization` header using HTTP Basic (default: `client_secret_post`) */
```
longsleep, line 46: This doc string could need better wording.

longsleep (approved):
This is awesome - just a minor comment from my side.

src/OidcClientSettings.ts
pamapa: yes, that was really not good, i have changed that now

pamapa: yes, that was really not good, i have changed that now

Issue #157, opened 2021-10-25
replace UrlUtils with native URL and URLSearchParams


pamapa:
- also replace duplicated code `url = window.location.href` in `signinRedirectCallback(url = window.location.href)` and `signoutRedirectCallback(url = window.location.href)` with `url?: string`

kherock:
I've begun working on this - I was wondering what we should do about the *SignInResponse functions which take `url` as a parameter. They can currently handle absolute URLs, relative URLs, or plain query strings.

Real-world usage should always pass an absolute URL, but all of our test cases use a string like `"state=state"`. My preference is that we require the full return URL to be passed in with v2.

pamapa:
This is the last jigsaw part for 2.0.0, would be cool to have it next week...

kherock:
I'm looking at cases where we default `url = window.location.href` and I'm not sure if it's a good idea to make it optional for the public methods:

* `signinCallback`/`signoutCallback` need it to read the `state` param to determine which navigator to use
* `signinRedirectCallback`/`signinPopupCallback`/`signinSilentCallback` are all part of the public API, and these methods **require** a url to function when they call `OidcClient#readSigninResponseState`. I think the `SigninResponse`/`SignoutReponse` constructors are mistyped in that the `url` argument should be a required param. Attempting to pass an empty url in those places will inevitably lead to runtime errors.

PR #194, opened 2021-11-06
Refactor navigators

kherock:
This cleans up navigator code. There might be some subtle differences from how things worked before, but I feel confident that this new implementation is much more robust. It has a couple of important improvements:

* callbacks attempt to compare to the `MessageEvent.source` property rather that the state ID. Not all browsers support this feature, so it still falls back to comparing to the state ID of the current navigation. This allows the navigator to recover even when the state param is entirely missing from the iframe/popup callback page.
* message payloads are encoded with a `source` key to prevent other libraries from interfering with navigator callbacks. The old code was deserializing with `JSON.parse` which could cause some console spam triggered by React dev tools while a popup window is open.

This refactoring should make #157's diff significantly cleaner.

### DIFF
--- a/.proxyrc.js
+++ b/.proxyrc.js
@@ -7,10 +7,9 @@ module.exports = function (app) {
         })
     );
 
-    app.use((req, res, next) => {
+    app.use('/code-flow-identityserver', (req, res, next) => {
         res.setHeader('Cross-Origin-Opener-Policy', 'unsafe-none');
         res.setHeader('Cross-Origin-Embedder-Policy', 'unsafe-none');
-		next();
-	});
-
+        next();
+    });
 };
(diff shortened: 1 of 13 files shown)


kherock:
I'm having trouble verifying these changes since it looks like the Parcel sample's popup navigation flow is broken. It's not clear to me why, but the `window.opener` context is lost as soon as navigation begins. If I change this line

https://github.com/authts/oidc-client-ts/blob/1d5f87aca139677ac356084505de0dee43def60e/samples/Parcel/oidc.js#L93

to `res.sendStatus(200)`, context is still lost. This is happening in Firefox, Chrome, and Edge. Safari 15 works however.

kherock:
Looks like it's a COOP/COEP issue. If I return these headers to their default values in `.proxyrc.js`, things work as expected.

pamapa (approved):

pamapa:
awesome work!

PascalLuginbuehl (approved):

PR #195, opened 2021-11-06
Replace UrlUtils with WhatWG URL APIs

### LINKED ISSUE #157: replace UrlUtils with native URL and URLSearchParams

kherock:
Closes #157

This refactors navigator code a bit further:

* URL parsing is now dictated solely by the settings of the window opener/iframe parent. The previous implementation extracted the `state` param from the URL on the child window and passed it separately from the full URL in `.postMessage()`.  Essentially this means that before, when using `response_mode=fragment`, the child window `settings` must match its parent since the parent window parses the URL according to its own `response_mode` setting after the navigator completes.

* `_signinCallback()` was inlined since the refactoring above made it into a trivial one-line function. Additionally, `signinPopupCallback` was lacking parity with `signoutPopupCallback`  w.r.t the `keepOpen` param.

### DIFF
--- a/src/SigninResponse.ts
+++ b/src/SigninResponse.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-import { Timer, UrlUtils } from "./utils";
+import { Timer } from "./utils";
 import type { UserProfile } from "./User";
 
 const OidcScope = "openid";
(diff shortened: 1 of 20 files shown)


pamapa:
awesome work!

src/SigninResponse.ts
```diff
@@ -29,31 +29,29 @@ export class SigninResponse {
...
+    public profile: UserProfile = {};
+
+    public constructor(params: URLSearchParams) {
+        // URLSearchParams returns `null` for missing values, reconstruct it as a map
+        const values = new Map(params);
+        this.error = values.get("error");
+        this.error_description = values.get("error_description");
+        this.error_uri = values.get("error_uri");
+
+        this.code = values.get("code")!;
```
kherock, line 44: Should I refactor this to throw an error if these required params aren't found? Or should I add exceptions for the lint warning?

pamapa: Technically it would be possible, practically it wont happen.

Analog to other code parts i would throw an error, its better than suppress lint warnings

pamapa: i saw you added eslint suppress comments. Options i see:
- eslint suppress
- values.get("code") || "" (probably the best)
- throw error (very explicit, but hard)

src/SigninResponse.ts
```diff
@@ -29,31 +29,31 @@ export class SigninResponse {
...
+        this.token_type = values.get("token_type")!;
+        /* eslint-enable */
+
+        this.state_id = values.get("state");
+        this.id_token = values.get("id_token");
+        this.session_state = values.get("session_state");
+        this.refresh_token = values.get("refresh_token");
+        this.scope = values.get("scope");
+        const expiresIn = values.get("expires_in");
+        this.expires_in = expiresIn == null ? undefined : parseInt(expiresIn);
```
pamapa, line 56: expiresIn can not be null, but undefined
->
this.expires_in = parseInt(expiresIn || "");

parseInt("") gives undefined...

kherock: I didn't realize that `parseInt` had that behavior, I'll update this

kherock: `parseInt("")` actually yields `NaN` which I don't think is what we want. I think the code I have is the soundest option still (expiresIn is nullish -> leave undefined ; else parse as integer).

kherock:
@pamapa I was also wondering what your thoughts are on adding the [@codecov](https://github.com/apps/codecov) bot to this org? It's pretty easy to set up.

pamapa:
> 
> 
> @pamapa I was also wondering what your thoughts are on adding the [@codecov](https://github.com/apps/codecov) bot to this org? It's pretty easy to set up.

looks nice and part of github. Coverage is always a good thing...

src/SigninResponse.ts
pamapa: you are right the console tricked me :-( argh

still expiresIn can not be null, why not simple:
`this.expires_in = expiresIn ? parseInt(expiresIn) : undefined;`

kherock: Yeah that's fine actually, I forgot that we don't have to worry about properly handling the `0` case.

src/SigninResponse.ts
kherock: I looked into this a bit further, it looks like we depend on ResponseValidator to clean up the bad responses. I'll default them to empty strings since I also agree that it's the least hacky.

pamapa:
just tell me when its ready for final review + mergable

src/SigninResponse.ts
```diff
@@ -29,31 +29,29 @@ export class SigninResponse {
...
+
+    public constructor(params: URLSearchParams) {
+        // URLSearchParams returns `null` for missing values, reconstruct it as a map
+        const values = new Map(params);
+        this.error = values.get("error");
+        this.error_description = values.get("error_description");
+        this.error_uri = values.get("error_uri");
+
+        this.code = values.get("code") ?? "";
+        this.access_token = values.get("access_token") ?? "";
```
pamapa, line 45: for maintainability it probably makes sense to add here a comment before the code block with the `?? ""`. Just in case others are wonder why no throw. Something like:

`ResponseValidator takes care about invalid values`

kherock:
This is ready to merge!

pamapa (approved):

PR #199, opened 2021-11-08
Automatically center popups in front of the opener window

kherock:
This updates the popup window navigator to automatically open and place the window centered in front the window opener. This is accomplished by refactoring the `popupWindowFeatures` setting to be an object instead of a string and introducing a new `PopupUtils` library which can compute a centered placement based on window dimensions.

Screenshot:
image

### DIFF
--- a/src/UserManagerSettings.ts
+++ b/src/UserManagerSettings.ts
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
 import { OidcClientSettings, OidcClientSettingsStore } from "./OidcClientSettings";
+import type { PopupWindowFeatures } from "./utils";
 import { WebStorageStateStore } from "./WebStorageStateStore";
 
 const DefaultAccessTokenExpiringNotificationTimeInSeconds = 60;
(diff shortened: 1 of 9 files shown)


kherock:
I'm pretty happy with the existing behavior, but I think it might also be nice to make the default width responsive based on the parent window width:

min width: `360px` (dense mobile layout)
&GreaterEqual; `480px`: `480px`
&GreaterEqual; `600px`: `600px` (should consistently trigger a tablet UI)

pamapa:
this one needs a git rebase for easier reviewing

src/UserManagerSettings.ts
```diff
@@ -15,10 +16,11 @@ export interface UserManagerSettings extends OidcClientSettings {
     popup_redirect_uri?: string;
     popup_post_logout_redirect_uri?: string;
     /**
-     * The features parameter to window.open for the popup signin window.
-     * (default: "location=no,toolbar=no,width=500,height=500,left=100,top=100;")
+     * The features parameter to window.open for the popup signin window. By default, the popup is
+     * placed centered in front of the window opener.
+     * (default: "location=no,toolbar=no,width=576,height=640")
```
pamapa: comment needs to be adapted as it is an object now:
(default: { location: "no", .... })

src/UserManagerSettings.ts
```diff
@@ -18,7 +18,6 @@ export interface UserManagerSettings extends OidcClientSettings {
     /**
      * The features parameter to window.open for the popup signin window. By default, the popup is
      * placed centered in front of the window opener.
-     * (default: "location=no,toolbar=no,width=576,height=640")
```
pamapa: why is the default removed, in the code we still use a default...
+
comment needs to be adapted as it is an object now:
(default: { location: false, .... })

kherock:
> this one needs a git rebase for easier reviewing

Sorry, I meant to do this when I marked this ready for review but ended up accidentally pushing an older commit instead.

pamapa:
looks great, will merge when pipeline succeeds

pamapa (approved):

pamapa:
@kherock I will create a 2.0.0-rc1 this week, i assume the API is stable now...

PR #213, opened 2021-11-19
Updates postForm to return JSON error message

rhigdon:
Returns proper json message in exception


rhigdon:
I am not sure this is the correct fix.  It would be nice if an Error shape was returned in the form of an oidc error.

ie. Keycloak returns `{ 'error': 'message', 'error_description': 'description of error'}`

pamapa (approved):

pamapa:
nice catch, that is a bug from refactoring, thanks for providing a fix

PR #215, opened 2021-11-20
improve unit-tests for JsonService

### LINKED ISSUE #214: improve unit-tests for JsonService
JsonService.postForm is currently untested, not good for maintenance and we had a bug #213, which is fixed now

pamapa:
closes #214

i am using `"test": "tsc -p test/unit/tsconfig.json && jest --coverage",` to detect missing spots, we have quite some...

### DIFF
--- a/test/unit/JsonService.test.ts
+++ b/test/unit/JsonService.test.ts
@@ -19,41 +19,33 @@ describe("JsonService", () => {
     });
 
     describe("getJson", () => {
-
-        it("should return a promise", async () => {
+        it("should reject promise when no url is passed", async () => {
             // act
-            const p = subject.getJson("http://test");
-
-            // assert
-            expect(p).toBeInstanceOf(Promise);
-            // eslint-disable-next-line no-empty
-            try { await p; } catch {}
+            await expect(subject.getJson(""))
+                // assert
+                .rejects.toThrow("url");
         });
 
         it("should make GET request to url", async () => {
             // act
-            const p = subject.getJson("http://test");
+            await expect(subject.getJson("http://test")).rejects.toThrow();
 
             // assert
             expect(fetchMock).toBeCalledWith("http://test", {
                 headers: {},
                 method: "GET"
             });
-            // eslint-disable-next-line no-empty
-            try { await p; } catch {}
         });
 
         it("should set token as authorization header", async () => {
             // act
-            const p = subject.getJson("http://test", "token");
+            await expect(subject.getJson("http://test", "token")).rejects.toThrow();
 
             // assert
             expect(fetchMock).toBeCalledWith("http://test", {
                 headers: { Authorization: "Bearer token" },
                 method: "GET"
             });
-            // eslint-disable-next-line no-empty
-            try { await p; } catch {}
         });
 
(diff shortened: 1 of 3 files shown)


test/unit/JsonService.test.ts
```diff
@@ -30,30 +29,41 @@ describe("JsonService", () => {
...
-            try { await p; } catch {}
+            try {
+                // act
+                await subject.getJson("http://test", "token");
+            } catch {
+                // assert
+                expect(fetchMock).toBeCalledWith("http://test", {
+                    headers: { Authorization: "Bearer token" },
+                    method: "GET"
+                });
```
kherock, line 65: This test will create a false positive if `.getJson` unexpectedly succeeds. Use something like this for testing async errors:

```suggestion
                await expect(subject.getJson("http://test", "token")).rejects.toThrow(Error);
                expect(fetchMock).toBeCalledWith("http://test", {
                    headers: { Authorization: "Bearer token" },
                    method: "GET"
                });

File: test/unit/JsonService.test.ts [hunk 2]
```diff
@@ -74,6 +84,28 @@ describe("JsonService", () => {
...
+            try {
+                // act
+                await subject.getJson("http://test");
+                fail("should not come here");
+            }
+            catch (err) {
+                // assert
+                expect(err).toBeInstanceOf(SyntaxError);
+                expect((err as SyntaxError).message).toContain("Unexpected token a in JSON");
+            }
```
kherock, line 106: Same here, you can reduce this to

```suggestion
            await expect(subject.getJson("http://test")).rejects.toThrow(new SyntaxError("Unexpected token a in JSON"));
```

test/unit/JsonService.test.ts
```diff
@@ -148,5 +182,314 @@ describe("JsonService", () => {
...
+    describe("postForm", () => {
+        it("should return a promise", async () => {
+            // act
+            const p = subject.postForm("http://test", { "a": "b" });
+
+            // assert
+            expect(p).toBeInstanceOf(Promise);
+            // eslint-disable-next-line no-empty
+            try { await p; } catch {}
+        });
```
kherock, line 216: do we need tests like this? They don't add to the line coverage and TypeScript already enforces type safety.

test/unit/JsonService.test.ts
```diff
@@ -148,5 +182,314 @@ describe("JsonService", () => {
...
+                expect(err).toBeInstanceOf(Error);
+                expect((err as Error).message).toContain("url");
+            }
+        });
+
+        it("should make POST request to url", async () => {
+            try {
+                // act
+                await subject.postForm("http://test", { "a": "b" });
+            } catch {
```
kherock, line 233: It seems like a lot of these `try/catch` patterns that only test if the request was attempted should actually be `try/finally`s.

pamapa:
I have now eliminate the error prone try/catch in the two test files with the pattern you suggested. Also the check for promise is now gone.

kherock (approved):
Awesome these look good now. Also just a heads up, I have some tests improving the coverage of UserManager and PopupWindow that I'll check in once the coverage diffs are up

PR #216, opened 2021-11-20
build: add codecov uploader job

kherock:
This splits the `pull_request` action into two jobs
* `build` validates that the code is theoretically shippable
* `test` checks for lint errors and uploads test coverage

Additionally, I've enabled these jobs for the `main` branch so that codecov can calculate the diff coverage correctly.


kherock:
FYI codecov's PR links will be broken until it's able to generate diffs from `main`'s coverage. Here's the "browse" view for one of the commits in this branch:

pamapa:
good point to enable codecov, now we can close the test gaps...

jest.config.mjs
```diff
@@ -1,8 +1,14 @@
-module.exports = {
+import yn from "yn";
+
+const collectCoverage = yn(process.env.CI);
```
pamapa, line 3: do we really need here a new dev dependency?. Alternate solution `const collectCoverage = process.env.CI ?? false` might work too?

kherock: I prefer this because I want to properly handle the `CI=false` and `CI=0` cases.

pamapa (approved):

kherock:
@pamapa since this changes the commit status names, we'll need to update the required statuses for merging to include the new names.

PR #217, opened 2021-11-20
Move unit tests into `src` folder

kherock:
This project has a lot of files to the extent that it's difficult to tell when files are missing an accompanying unit test suite. Additionally, it's not ideal to have to backtrack up several directories (`../../src/foo/bar`) for every import in each test suite.

This refactors our tsconfigs into `tsconfig.json` and `tsconfig.build.json`. The former is loaded by your IDE and performs type-checking on the whole repo before running tests, and the latter is used by API Extractor to generate the type definition bundle.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -56,6 +56,9 @@ overrides:
   - files: src/**/*
     env:
       node: false
-  - files: test/**/*
+  - files: [src/**/*.test.*, test/**/*]
     env:
       jest: true
+    rules:
+      "@typescript-eslint/unbound-method": "off"
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           path: |
             lib/
-            tsconfig.tsbuildinfo
+            tsconfig.build.tsbuildinfo
           key: build-${{ github.head_ref }}-${{ github.sha }}
           restore-keys: |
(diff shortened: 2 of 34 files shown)


pamapa (approved):

pamapa:
makes sense, nice work

pamapa:
ready to merge?

pamapa (approved):

kherock:
@pamapa I'll merge in a minute, was rebasing, and I just noticed that I need to update the `ci.yml` to reflect the new `.tsbuildinfo` locations.

pamapa:
ok, i let you merge this

pamapa:
regarding CI, with `CI.yml`: `release.yml` is somehow redundant and can be removed now, as CI.yml now also takes care of main...

kherock:
ah I didn't realize that `release` didn't have side effects. Naming-wise I think it might make sense to get rid of it and rename `publish.yml` &rarr; `release.yml`.

pamapa:
If i make a release by hand in github that workflow triggers. Originally i have chosen the name `publish.yml`, as it publishes that new release to the npm registry. But maybe your are right...

kherock:
Yeah I usually name the workflow yml as a description of its triggers, and then the jobs themselves describe the task they accomplish. so having `release.yml` contain a job named "publish" that does what you described is how I approach it.

PR #218, opened 2021-11-21
Refactor json service

kherock:
This properly resolves the issue addressed in #213 by throwing an `ErrorResponse` instance when the response is parseable as JSON and contains a `.error` value.

The unit test code checked into #215 will need some minor updates if this merges first.


pamapa:
The JsonService commits look good, can you cherry pick them onto a separate merge request. The other two commits seem to be part of two other merge requests,....

pamapa:
nice work, simplifies code quite a bit

pamapa (approved):

PR #220, opened 2021-11-21
Update tests for PopupWindow and UserManager

kherock:
Adds some error handling tests for PopupWindow's and some smoke tests UserManager's redirect flows.

I'll probably be working on smoke tests for the rest of UserManager's public API next.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -60,5 +60,6 @@ overrides:
     env:
       jest: true
     rules:
+      "@typescript-eslint/no-non-null-assertion": "off"
       "@typescript-eslint/unbound-method": "off"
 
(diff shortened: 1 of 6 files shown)


pamapa:
lint fails

pamapa (approved):

kherock:
@pamapa thoughts on disabling `@typescript-eslint/no-non-null-assertion` for test files? The rule essentially exists toprotect you from runtime errors. In test files, I can always can interpret `!` to mean "this should trigger a runtime error and fail the test when this property is unexpectedly nullish".

pamapa:
Nice work, now the goal is to make codecov go higher.
It is late here, i will write tomorrow some more tests, e.g. for ResponseValidator.ts

PR #227, opened 2021-11-25
fix 225

pamapa:
- extend Parcel sample with revoke button for testing
- make `revokeAccessToken` work at all: The `if (!access_token || access_token.includes(".")) {` prevented it, as valid access token include ".", as we dropped implicit flow we can remove that guard anyway.
- move revoke into ClientToken and align function signature to other functions inside this class
- added unit-test for ClientToken


src/JsonService.ts
```diff
@@ -99,22 +99,27 @@ export class JsonService {
...
-            this._logger.error("postForm: Error parsing JSON response", err);
-            if (response.ok) throw err;
-            throw new Error(`${response.statusText} (${response.status})`);
+
+        let json: Record<string, unknown> = {};
+        if (hasResponseData) {
+            try {
+                json = await response.json();
+            }
+            catch (err) {
```
kherock, line 108: I'm not a fan of this new parameter, I think this function should be refactored to optimistically parse a non-empty response body instead. Something like this:

```suggestion
        const body = await response.text();
        let json: Record<string, unknown> = {};
        if (body) {
          try {
            json = JSON.parse(body);
          }
          catch { ... }
        }
```

pamapa: good point, i will adapt, should we add a try catch around response.text() or can't that happen?

kherock: I'm not aware of a situation where it can throw outside of poor network conditions

src/MetadataService.ts
```diff
@@ -100,8 +100,8 @@ export class MetadataService {
         return this._getMetadataProperty("end_session_endpoint", true) as Promise<string | undefined>;
     }
 
-    public getRevocationEndpoint(): Promise<string | undefined> {
-        return this._getMetadataProperty("revocation_endpoint", true) as Promise<string | undefined>;
+    public getRevocationEndpoint(optional = true): Promise<string | undefined> {
```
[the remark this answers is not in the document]
pamapa: i agree, but lets not mix this with this merge request, lets do that after this is merged.
BTW:  getRevocationEndpoint is called with optional as variable argument and not as const...

pamapa:
ready for review/merging again

kherock (approved):

PR #236, opened 2021-12-03
fix signout callback

pamapa:
- the signoutCallback always throws otherwise an error
- found issue by writing tests

### DIFF
--- a/src/UserManager.ts
+++ b/src/UserManager.ts
@@ -321,32 +321,35 @@ export class UserManager {
         this._logger.info("signinSilentCallback: successful");
     }
 
-    public async signinCallback(url = window.location.href): Promise<User | null> {
+    public async signinCallback(url = window.location.href): Promise<User | void> {
         const { state } = await this._client.readSigninResponseState(url);
-        if (state.request_type === "si:r") {
-            return this.signinRedirectCallback(url);
+        switch (state.request_type) {
+            case "si:r":
+                return this.signinRedirectCallback(url);
+            case "si:p":
+                return await this.signinPopupCallback(url);
+            case "si:s":
+                return await this.signinSilentCallback(url);
+            default:
+                throw new Error("invalid response_type in state");
         }
-        if (state.request_type === "si:p") {
-            await this.signinPopupCallback(url);
-            return null;
-        }
-        if (state.request_type === "si:s") {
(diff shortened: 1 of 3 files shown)


pamapa:
@kherock when this is merged i would like to make the 2.0.0 release, what do you think?

src/UserManager.ts
```diff
@@ -323,30 +323,35 @@ export class UserManager {
...
     public async signinCallback(url = window.location.href): Promise<User | null> {
         const { state } = await this._client.readSigninResponseState(url);
-        if (state.request_type === "si:r") {
-            return this.signinRedirectCallback(url);
+        switch (state.request_type) {
+            case "si:r":
+                return this.signinRedirectCallback(url);
+            case "si:p":
+                await this.signinPopupCallback(url);
+                return null;
```
kherock, line 331: I think the `null` return value on this function is a bit strange given that all of the other `*Callback` methods use `void` as their return type. Could we update this function to have a `Promise<User | void>` return value and simplify these to 

```suggestion
                return await this.signinPopupCallback(url);
```
?

kherock (approved):
This looks good, sorry that I never got around to writing those UserManager tests like I said I would.

As for the v2 release, I would like to take a moment to review our eslint rules one more time and potentially enforce [`@typescript-eslint/return-await`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/return-await.md) for better async stack traces. After that, I think this project is pretty stable. It has all of the quality of life enhancements I was wishing the v1 library had!

src/UserManager.ts
pamapa: hmm `signinPopupCallback` calls `this._popupNavigator.callback`, which returns void. I have looked into the callback method, no glue were to get the user from...
Any way this is something else than what this merge request fixes. Can you provide an improvement for that?

kherock: Yeah there's no way to get a user from the popup/iframe callbacks, the `void` return is appropriate. I only figured this change is relevant here since you're making tests that enforce returning `null`.

pamapa: ahh got it `void` vs `null`. Will adapt...

pamapa:
fixed the null vs void thing

Issue #125, opened 2021-10-07
UserProfile additional claims from IDP


OneCyrus:
the returned token from our IdP includes some additional claims like client_id and name. the current implementation has a strict UserProfile interface which doesn't allow to access this data.

It would be great to access additional claims. May be we could have a generic way to access them through an extended UserProfile type.

pamapa:
good point, feel free to provide a merge request.
claims which are in the OAuth/OIDC standard should be easy accessible...

PR #241, opened 2021-12-06
feat(User): add types for standard claims to UserProfile

pbklink:
Firstly, thank-you for taking on maintenance of oidc-client (and migrating to TypeScript)

As per #125, this PR adds extra properties to UserProfile so that standard claims can be accessed by applications.

Since the code already existed to merge these claims into UserProfile (`ResponseValidator._mergeClaims()`), all this PR does is add the extra properties so that they can be accessed with TypeScript applications.

### DIFF
--- a/src/User.ts
+++ b/src/User.ts
@@ -12,6 +12,27 @@ export interface UserProfile {
     azp?: string;
     at_hash?: string;
     auth_time?: number;
+
+    // Other standard claims - https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+    name?: string;
+    given_name?: string;
+    family_name?: string;
+    middle_name?: string;
+    nickname?: string;
+    preferred_username?: string;
+    profile?: string;
+    picture?: string;
+    website?: string;
+    email?: string;
+    email_verified?: boolean;
+    gender?: string;
+    birthdate?: string;
+    zoneinfo?: string;
+    locale?: string;
+    phone_number?: string;
+    phone_number_verified?: boolean;
+    address?: Record<string, unknown>; // https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim
+    updated_at?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/User.ts
```diff
@@ -12,6 +12,27 @@ export interface UserProfile {
     azp?: string;
     at_hash?: string;
     auth_time?: number;
+
+    // Other standard claims - https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
```
kherock, line 16: It would be helpful to include this link in the TSDoc for the class definition above (before the `@public` hint). Something like

```typescript
/**
  * @see https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
  *
  * @public
  */
export interface UserProfile {
```

src/User.ts
```diff
@@ -12,6 +12,27 @@ export interface UserProfile {
...
+    website?: string;
+    email?: string;
+    email_verified?: boolean;
+    gender?: string;
+    birthdate?: string;
+    zoneinfo?: string;
+    locale?: string;
+    phone_number?: string;
+    phone_number_verified?: boolean;
+    address?: Record<string, unknown>; // https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim
```
kherock, line 34: I'd again leverage TSDoc here:

```suggestion
    /** @see https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim */
    address?: Record<string, unknown>
```

Doing it this way allows for it to be picked up by the language server and in the Typedoc!

image

kherock (commented):
Thanks for the PR! Looks fine aside from the TSDoc suggestions.

pbklink:
TSDoc suggestions implemented.

Also added documentation for individual properties.  Information was copied directly from standard.

Let me know if you would like anything else changed.

src/User.ts
pbklink: See comment below

kherock (approved):
Awesome 👍

PR #246, opened 2021-12-08
Eslint updates

### LINKED ISSUE #243: enforce @typescript-eslint/return-await for better async stack traces

kherock:
Resolves #243 and enforces a few more rules which should hopefully make sense:

* [`return-await`](https://typescript-eslint.io/rules/return-await/) - setting this to `always` ensures that async stack traces don't skip over functions that return promises. It also removes the ambiguity over when it's correct to `return await` inside of a try/catch.
* [`comma-dangle`](https://eslint.org/docs/rules/comma-dangle) - enables dangling commas for multiline functions and properties. This permits cleaner diffs when keys/arguments are appended or reordered
* [`no-multi-spaces`](https://eslint.org/docs/rules/no-multi-spaces) - this rule is sometimes controversial when you want to do fancy alignment across multiple lines where the space between operators is padded with whitespace to make things line up.

no-unsafe-call, no-unsafe-member-access, no-unsafe-return, and explicit-module-boundary-types were re-enabled by the `@typescript-eslint/recommended` preset since I was able to remove the remaining `any` types from the project.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -35,6 +35,7 @@ overrides:
       - plugin:@typescript-eslint/recommended-requiring-type-checking
     rules:
       indent: "off"
+      no-return-await: "off"
       "@typescript-eslint/indent":
         - error
         - 4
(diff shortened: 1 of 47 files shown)


pamapa:
great work, that will help in the long run

pamapa (approved):

pamapa:
another rc or 2.0.0?

PR #264, opened 2021-12-15
Additional `window` updates

kherock:
This removes more unnecessary usage of `window`. The most important change here is that `window.crypto.getRandomValues()` is replaced by `CryptoJS.lib.WordArray.random()`. Under the hood, CryptoJS uses `window.crypto` but falls back to `window.msCrypto` (IE11) and `require('crypto')` (Node.js).


pamapa (approved):

pamapa:
great work and less code to maintain in the future.

PR #266, opened 2021-12-15
feat: return destructor functions from event listener methods

kherock:
This updates the `addAccessTokenExpired`, `addUserLoaded`, etc. methods to forward the destructor function returned by the underlying `Events` class. This is a nice pattern to support since React users can integrate it into a useEffect hook:

```javascript
useEffect(() => {
  return userManager.events.addAccessTokenExpiring(() => setSessionExpiring(true));
}, [userManager.events])


pamapa (approved):

pamapa:
nice work

PR #267, opened 2021-12-16
Refactor revoke token

### LINKED ISSUE #262: Revoking access token fails
Migrating from [oidc-client-js](https://github.com/IdentityModel/oidc-client-js/).
In `oidc-client-ts` revoking no longer works.

It is my understanding that when revoking an access token, it's the **refresh token** that should be sent to the revoke endpoint.
See the Cognito documentation:

> The refresh token that the client wants to revoke. All access tokens issued from this refresh token are also revoked.

https://docs.aws.amazon.com/cognito/latest/developerguide/revocation-endpoint.html

`oidc-client-ts` (wrongly?) sends the access token, which results in the following error:
> HTTP 400 and error unsupported_token_type is returned if the token sent in the revocation request is not a refresh token

kherock:
Resolves #262.

There's two questions that I couldn't really answer for myself while refactoring this:

1. Why does revoking tokens mutate the User? I see that since 1.x, `access_token`, `refresh_token`, `token_type`, etc. were updated to be set to empty strings after revocation rather than `null`. I don't see any benefit to doing this client-side since it would lead to unnecessarily opaque errors. If I make an API request after revoking a token, I'd rather be alerted by the API that my token is revoked and no longer valid rather than it rejecting it because I'm suddenly sending an empty string for an access token. I can see an argument for removing `refresh_token`, but everything else seems like it should stay put until the user signs out.

2. Why is revocation success optional on sign out? `revokeTokensOnSignout` needs to be explicitly opted into, so it feels sloppy to not alert the application when revocation fails there.

### DIFF
--- a/src/UserManager.ts
+++ b/src/UserManager.ts
@@ -28,6 +28,11 @@ export type ExtraSignoutRequestArgs = Pick<CreateSignoutRequestArgs, "extraQuery
 /**
  * @public
  */
+export type RevokeTokensTypes = UserManagerSettings["revokeTokenTypes"]
+
+/**
+ * @public
+ */
 export type SigninRedirectArgs = RedirectParams & ExtraSigninRequestArgs;
 
 /**
@@ -529,8 +534,8 @@ export class UserManager {
             const user = await this._loadUser();
             this._logger.debug("_signoutStart: loaded current user from storage");
 
-            if (this.settings.revokeAccessTokenOnSignout) {
-                await this._revokeInternal(user, true);
+            if (this.settings.revokeTokensOnSignout) {
+                await this._revokeInternal(user);
             }
 
             const id_token = args.id_token_hint || user && user.id_token;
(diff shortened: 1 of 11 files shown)


src/UserManager.ts
```diff
@@ -563,50 +563,30 @@ export class UserManager {
         return signoutResponse;
     }
 
-    public async revokeAccessToken(): Promise<void> {
+    public async revokeTokens(types = this.settings.revokeTokenTypes): Promise<void> {
```
pamapa, line 566: in order to keep consumer libraries simple, lets add here a type, like the other public functions have: `RevokeTokensArgs`

pamapa:
awesome you just beat me, i wanted to start working today on this and just saw that you are nearly done already, that is really cool!

src/UserManager.ts
```diff
@@ -563,50 +565,36 @@ export class UserManager {
         return signoutResponse;
     }
 
-    public async revokeAccessToken(): Promise<void> {
+    public async revokeTokens(types?: RevokeTokensArgs): Promise<void> {
```
pamapa, line 568: cosmetics: we could name types here `args` like in all other public functions

src/UserManager.ts
```diff
@@ -25,6 +25,8 @@ export type ExtraSigninRequestArgs = Pick<CreateSigninRequestArgs, "extraQueryPa
  */
 export type ExtraSignoutRequestArgs = Pick<CreateSignoutRequestArgs, "extraQueryParams" | "state">
 
+export type RevokeTokensArgs = UserManagerSettings["revokeTokenTypes"]
```
pamapa, line 28: needs @public

pamapa:
would be nice to add a notice into `docs/migration.md` about the rename + new setting

pamapa:
@matthetherington-pa i would like to have this MR in the next rc release. It is already quiet ready and should come in in the next few days. Can you wait or do you desperately need a new rc release?

matthetherington-pa:
@pamapa yep at your leisure is fine. After today I'll only be continuing in the new year

Thank you 🙏

pamapa:
> @pamapa yep at your leisure is fine. After today I'll only be continuing in the new year

@matthetherington-pa released for you :-)

src/UserManager.ts
kherock: To me `Args` implies a list of named params, I think I might prefer renaming the type to `RevokeTokensTypes` if that makes sense

pamapa (approved):

pamapa:
great work

PR #279, opened 2021-12-18
fix(PopupWindow): clear the interval checking popup closed on dispose

### LINKED ISSUE #274: "Window navigation aborted" event raised in an infinite loop
When performing a `signinPopup()` a "Window navigation aborted" event will be raised indefinitely.

Screen Shot 2021-12-16 at 19 23 06

The issue comes from [AbstractChildWindow.ts#L63](https://github.com/authts/oidc-client-ts/blob/72952f23e25ca4efe8c017662aaeb2ec228a8f39/src/navigators/AbstractChildWindow.ts#L63).

When disposing, the event is raised but it's also asking for disposal -> and here comes the loop.

Screen Shot 2021-12-18 at 12 01 20

kherock:
Fixes #274. The old timer would only clean up after itself in the event that navigation was aborted (when `this._abort.raise()` is called). This issue is compounded by the fact that the `dispose()` was removing `_abort`'s ability to run clean up logic since it removed the only relevant event listener.

The fix is just to have `dispose()` clear the interval directly.

### DIFF
--- a/src/navigators/PopupWindow.test.ts
+++ b/src/navigators/PopupWindow.test.ts
@@ -21,6 +21,7 @@ describe("PopupWindow", () => {
             focus: jest.fn(),
             close: jest.fn(),
         } as unknown as WindowProxy));
+        jest.useFakeTimers();
     });
 
     afterEach(() => {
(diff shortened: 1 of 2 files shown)


pamapa:
the unit-tests are failing, the fix looks good

pamapa (approved):

pamapa:
easy fix, nice

PR #280, opened 2021-12-20
feat(claims): improve claims types

Badisi:
@pamapa, sorry for the last PR. I went too fast on that one.

Here is a cleaner one and test should pass 😉

### DIFF
--- a/src/Claims.ts
+++ b/src/Claims.ts
@@ -1,0 +1,127 @@
+/**
+ * @internal
+ */
+export type Mandatory<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T, K>;
+
+/**
+ * Standard OpenID Connect address claim.
+ * The Address Claim represents a physical mailing address.
+ *
+ * @public
+ * @see https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim
+ */
+export interface OidcAddressClaim {
+    /** Full mailing address, formatted for display or use on a mailing label. This field MAY contain multiple lines, separated by newlines. Newlines can be represented either as a carriage return/line feed pair ("\\r\\n") or as a single line feed character ("\\n"). */
+    formatted?: string;
+    /** Full street address component, which MAY include house number, street name, Post Office Box, and multi-line extended street address information. This field MAY contain multiple lines, separated by newlines. Newlines can be represented either as a carriage return/line feed pair ("\\r\\n") or as a single line feed character ("\\n"). */
+    street_address?: string;
+    /** City or locality component. */
+    locality?: string;
+    /** State, province, prefecture, or region component. */
+    region?: string;
+    /** Zip code or postal code component. */
+    postal_code?: string;
+    /** Country name component. */
+    country?: string;
+}
+
+/**
+ * Standard OpenID Connect claims.
+ * They can be requested to be returned either in the UserInfo Response or in the ID Token.
+ *
+ * @public
+ * @see https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+ */
+export interface OidcStandardClaims {
+    /** Subject - Identifier for the End-User at the Issuer. */
+    sub?: string;
+    /** End-User's full name in displayable form including all name parts, possibly including titles and suffixes, ordered according to the End-User's locale and preferences. */
+    name?: string;
+    /** Given name(s) or first name(s) of the End-User. Note that in some cultures, people can have multiple given names; all can be present, with the names being separated by space characters. */
+    given_name?: string;
+    /** Surname(s) or last name(s) of the End-User. Note that in some cultures, people can have multiple family names or no family name; all can be present, with the names being separated by space characters. */
+    family_name?: string;
+    /** Middle name(s) of the End-User. Note that in some cultures, people can have multiple middle names; all can be present, with the names being separated by space characters. Also note that in some cultures, middle names are not used. */
+    middle_name?: string;
+    /** Casual name of the End-User that may or may not be the same as the given_name. For instance, a nickname value of Mike might be returned alongside a given_name value of Michael. */
+    nickname?: string;
+    /** Shorthand name by which the End-User wishes to be referred to at the RP, such as janedoe or j.doe. This value MAY be any valid JSON string including special characters such as \@, /, or whitespace. */
+    preferred_username?: string;
+    /** URL of the End-User's profile page. The contents of this Web page SHOULD be about the End-User. */
+    profile?: string;
+    /** URL of the End-User's profile picture. This URL MUST refer to an image file (for example, a PNG, JPEG, or GIF image file), rather than to a Web page containing an image. Note that this URL SHOULD specifically reference a profile photo of the End-User suitable for displaying when describing the End-User, rather than an arbitrary photo taken by the End-User. */
+    picture?: string;
+    /** URL of the End-User's Web page or blog. This Web page SHOULD contain information published by the End-User or an organization that the End-User is affiliated with. */
+    website?: string;
+    /** End-User's preferred e-mail address. Its value MUST conform to the RFC 5322 addr-spec syntax. */
+    email?: string;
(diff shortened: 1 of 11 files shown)


src/Claims.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Standard OpenID Connect claims.
+ * They can be requested to be returned either in the UserInfo Response or in the ID Token.
+ *
+ * @internal
+ * @see https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+ */
+export interface OidcClaims {
+    [customClaimKey: string]: unknown;
```
kherock, line 31: I think `OidcStandardClaims` would be a better name for this - to me "ID Token claims" are synonymous with "OIDC claims", so it's a bit confusing to see this interface not contain the other required claims like "sub", "aud", etc.

Additionally, this `customClaimKey` index type feels out of place here especially since you already redefine in the `IdTokenClaims` interface.

src/Claims.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/**
+ * Standard JWT claims.
+ *
+ * @internal
+ * @see https://datatracker.ietf.org/doc/html/rfc7519#section-4.1
+ */
+export interface JwtClaims {
+    [customClaimKey: string]: unknown;
```
kherock, line 82: Nitpick, I think this index type should be named something like `otherClaim` or just `claim` since additional claims by the JWT spec are often public, registered names (such as those used for OIDC).

src/Claims.ts
```diff
@@ -0,0 +1,124 @@
...
+    jti?: string;
+}
+
+/**
+ * Standard ID Token claims.
+ *
+ * @internal
+ * @see https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+ */
+export interface IdTokenClaims extends JwtClaims {
```
kherock, line 106: This should also extend standard claims interface above.

Also, I liked that Mandatory type you had on the JwtClaims from the last PR! I wouldn't mind having that back.

kherock (commented):
This looks good, I appreciate the effort you went through to pull in the docs from the OIDC spec!

src/Claims.ts
Badisi: 1) Good catch, thanks
2) I brought back the Mandatory type but I had to refactor the tests to make it work though - let me know if the unknown cast is ok for you

src/Claims.ts
Badisi: Done

src/Claims.ts
Badisi: 1) Done
2) Agreed. Done

pamapa:
nice work, i let @kherock doing the final review an approval...

pamapa:
Please re-check commit and avoid stuff like `Merge branch` in MR. On MR you should operate with `git rebase` and it would be nice if you could squash all commits into one...

Badisi:
@pamapa, I had to find out about the `build-types` script to get the `temp/` file, but APIs doc is now up to date.

Sorry about the `Merge branch`, I used the `Fetch upstream` feature directly from `github` and it made it appear here.
As for the `squash`, can't you just squash them when accepting the PR ?
Thanks

pamapa:
@Badisi When we squash the commit author changes...

Badisi:
@pamapa, I was not aware of that issue..

All my commits for that PR are now squashed.

pamapa:
Besides of the single commit there are still 3 commits inside, which were made by kherock and already landed on main. You can easily get ride of them with a git rebase against remote main...

Badisi:
Done, sorry for the inconvenience.

src/ResponseValidator.test.ts
```diff
@@ -453,7 +453,7 @@ describe("ResponseValidator", () => {
     describe("_mergeClaims", () => {
         it("should merge claims", () => {
             // arrange
-            const c1 = { a: "apple", b: "banana" } as UserProfile;
+            const c1 = { a: "apple", b: "banana" } as unknown as UserProfile;
```
kherock, line 456: This cast is fine for now, a more proper fix will probably be to make `_mergeClaims` generic with something like `_mergeClaims<T extends JwtClaims>(claims1: T, ...): T`. Thanks for going through all of these tests.

src/User.ts
```diff
@@ -2,65 +2,13 @@
...
-     *
-     * @see https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim
-     */
-    address?: Record<string, unknown>;
-    /** Time the End-User's information was last updated. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time. */
-    updated_at?: number;
-
-    [claim: string]: unknown;
-}
+export type UserProfile = IdTokenClaims & Mandatory<OidcStandardClaims, "sub">;
```
kherock, line 11: The OidcStandardClaims is redundant since IdTokenClaims already extends from it and makes `sub` required.

Additionally I'd prefer keeping this as an interface type even if it's empty

```suggestion
export interface UserProfile extends IdTokenClaims {}
````

File: src/Claims.ts [hunk 6]
```diff
@@ -0,0 +1,127 @@
+/**
+ * @public
+ */
+export type Mandatory<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T, K>;
```
kherock, line 4: The `export * from './Claims'` from the library entrypoint causes this to be exported as part of the public API. GIven that `UserProfile` shouldn't need this type with the above comment, I suggest keeping this type local to this file.

kherock (commented):
Nearly there, thanks for addressing everything.

@pamapa fyi, squashing from the github UI generally does preserve author data provided that every commit in the branch is by the same author. Obviously in this case my upstream commits that were replayed with @Badisi as the committer would have caused problems since I would have become co-author of the squash.

brockallen:
Does this change anything about the `User` or the `profile` that holds the claims? Or are these just constants to use in the id_token processing?

src/User.ts
Badisi: 1. Agreed
2. This is going to make the linter complain : 
> An interface declaring no members is equivalent to its supertype
> eslint@typescript-eslint/no-empty-interface

src/Claims.ts
Badisi: `UserProfile` needs to be exported, and as it extends from `IdTokenClaims` so does `IdTokenClaims`, `Madatory`, `OidcStandardClaims` and `JwtClaims`.

Otherwise the following would appear in the APIs doc:
```md
// Warning: (ae-forgotten-export) The symbol "Mandatory" needs to be exported by the entry point index.d.ts
// Warning: (ae-forgotten-export) The symbol "OidcStandardClaims" needs to be exported by the entry point index.d.ts
// Warning: (ae-forgotten-export) The symbol "JwtClaims" needs to be exported by the entry point index.d.ts
```

kherock: `export * from './Claims'` is fine, I just don't want a utility type like `Mandatory` being exported by this library. In this case the` (ae-forgotten-export) The symbol "Mandatory" needs to be exported by the entry point index.d.ts` is acceptable

kherock:
@brockallen These actually don't have runtime implications, they're just organizing the type definitions a bit bitter.

src/User.ts
kherock: Oh interesting, I didn't realize `allowSingleExtends` was turned off here. The `export type` is ok.

pamapa:
@Badisi  Conflicting files (src/ResponseValidator.ts). Please rebase against remote main, i have just merged a refactor from kherock...

Badisi:
@pamapa, rebase was done. Any other blocking points ?

pamapa:
@Badisi for me it looks fine.

@kherock please finish the review...

kherock (approved):
Sorry for the delay getting back to this. LGTM

pamapa:

PR #281, opened 2021-12-20
fix(#278): expose the internal logger

### LINKED ISSUE #278: expose the internal logger
I have implemented my own storage class for mobile purpose:
```ts
class MobileStorage implements Storage {
    ...
}

    userStore: new WebStorageStateStore({
        store: (isNative) ? new MobileStorage() : new InMemoryWebStorage()
    })
});
```

I would like to benefits from the internal Logger like here: [InMemoryWebStorage.ts#L14](https://github.com/authts/oidc-client-ts/blob/72952f23e25ca4efe8c017662aaeb2ec228a8f39/src/InMemoryWebStorage.ts#L14) but as it is internal I don't have access to it.

Any chance to make it public somehow ?

Badisi:
Fix #278

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -16,8 +16,6 @@ export class AccessTokenEvents {
     addAccessTokenExpiring(cb: AccessTokenCallback): () => void;
     // (undocumented)
     load(container: User): void;
-    // Warning: (ae-forgotten-export) The symbol "Logger" needs to be exported by the entry point index.d.ts
-    //
     // (undocumented)
     protected _logger: Logger;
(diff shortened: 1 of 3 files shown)


pamapa:
looks good!

Currently the build fails, because the APi changes, due to the improved user profile type:
`Please copy the file "temp/oidc-client-ts.api.md" to "docs/oidc-client-ts.api.md"` and add that file in the MR too....

pamapa:
i still only see one file changed, please re-check commit and avoid stuff like `Merge branch` in MR. On MR you should operate with `git rebase` and squash...

pamapa (approved):

pamapa:
nice work, thanks for contributing to this library!

PR #282, opened 2021-12-20
Update user profile after refresh token requests

### LINKED ISSUE #276: profile is not updated after silent refresh
Bringing back an old issue from `oidc-client-js` as it still exists: [#1034](https://github.com/IdentityModel/oidc-client-js/issues/1034).

### Issue

Create a page with two manual actions : `login` and `silent renew`.

1) Load the page and trigger a login
    ```ts
    await this.userManager.signinRedirect();
    ```
2) Open a new tab, go to your IDP and change the `First Name` of the user
3) Go back to the page and trigger a silent renew
    ```ts
    await this.userManager.signinSilent();
    ```
4) The request returned an updated `id_token` with `"given_name": "new name"`
5) But the `user.profile` is not refreshed

### Workaround

I'm currently using the following workaround :

```ts
interface IdToken {
    iss?: string;
    aud?: string | string[];
    azp?: string;
    iat?: number;
    nbf?: number;
    exp?: number;
    sub?: string;
    auth_time?: number;
};

const user = await this.userManager.signinSilent();
if (user?.id_token) {
    Object.assign(user.profile, jwtDecode<IdToken>(user.id_token));
    await this.userManager.storeUser(user);
}
```

kherock:
Resolves #276.

When planning this out, I would've like to simply reuse the logic declared in `ResponseValidator`, but its private methods (and unfortunately its tests) were coupled to the use of the `SigninState` object.

In this PR I've refactored `ResponseValidator` tests to only use the public methods. I excluded `_mergeClaims` and `_filterProtocolClaims` from this since those methods are essentially utility functions that probably should be public exports in their own file(s). This allowed me to more easily refactor  `_validateTokens` and `_processClaims` to work without signin state. This will be necessary for validating refresh token responses as they do not require the same kind of state.

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -563,7 +563,7 @@ export class SigninResponse {
     // (undocumented)
     access_token: string;
     // (undocumented)
-    readonly code: string;
+    readonly code: string | undefined;
     // (undocumented)
     error: string | undefined;
     // (undocumented)
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -228,9 +228,9 @@ export class OidcClient {
     // (undocumented)
     clearStaleState(): Promise<void>;
     // (undocumented)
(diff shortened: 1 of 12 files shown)


pamapa:
really great work awesome!

pamapa (approved):

PR #284, opened 2021-12-20
fix(UserManager): fix anonymous session monitoring

kherock:
Small bug I noticed where `querySessionStatus` would wrongly throw if the error object contained a message that was set from the `error_description` value rather than `error`.


pamapa:
cool, much cleaner.

pamapa (approved):

PR #289, opened 2021-12-21
Misc cleanup

kherock:
After implementing the refresh token fixes I have a much better understanding of the SigninResponse/ResponseValidator relationship and was able to clean up code a bit further.

The `SigninResponse` constructor was originally built to take in a list of URL params. For the implicit `token` grant, there are a bunch of these - `access_token`, `token_type`, `expires_in`, etc. In the `code` flow, there are only a handful - `code`, `state`, `session_state`, and the error params.

In this PR I've limited the `SigninResponse` constructor to only read properties expected for the `code` grant type. I've also renamed `state_id` back to `state` and `state` to `userState`. This doesn't affect the public User API since the `User` constructor maps `userState` to `state`.

Within all of these changes is also a really minor fix to preserve user state after performing a refresh token grant.

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -9,7 +9,7 @@ export type AccessTokenCallback = (...ev: unknown[]) => void;
 
 // @public (undocumented)
 export class AccessTokenEvents {
-    constructor({ expiringNotificationTimeInSeconds }: {
+    constructor(args: {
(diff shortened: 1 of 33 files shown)


pamapa:
Yes its more readable when we use internally `userState`! And the cleanup is also very nice. The library is now getting into a really nice shape!

pamapa (approved):

PR #291, opened 2021-12-22
refactor: additional logger improvements

kherock:
I wanted figure out a way to avoid specifying the function name for every logger call or duplicating error messages when logging an error before throwing and came up with a few more improvements:

* added `logger.create` which creates a short-lived clone of a logger object with a `_method` name set
  * upon creation, a debug statement `[<Class>] <method>: begin` is logged
* added `logger.throw` which logs an error object and throws it


src/ResponseValidator.ts
```diff
@@ -209,26 +203,21 @@ export class ResponseValidator {
     }
 
     protected _filterProtocolClaims(claims: UserProfile): UserProfile {
-        this._logger.debug("_filterProtocolClaims, incoming claims:", claims);
```
pamapa, line 212: at some point we should log the incoming claims, if not here then here
`logger.debug("response is OIDC, processing claims", response.profile);`?

src/SigninState.ts
```diff
@@ -73,7 +73,7 @@ export class SigninState extends State {
     }
 
     public toStorageString(): string {
-        Logger.debug("SigninState", "toStorageString");
+        new Logger("SigninState").create("toStorageString");
```
pamapa, line 76: will this log a debug message?

src/SigninState.ts
```diff
@@ -93,7 +93,7 @@ export class SigninState extends State {
     }
 
     public static fromStorageString(storageString: string): SigninState {
-        Logger.debug("SigninState", "fromStorageString");
+        Logger.debug("SigninState.fromStorageString");
```
pamapa, line 96: new Logger("SigninState").create("fromStorageString");

src/State.ts
```diff
@@ -44,15 +44,15 @@ export class State {
     }
 
     public static fromStorageString(storageString: string): State {
-        Logger.debug("State", "fromStorageString");
+        Logger.debug("State.fromStorageString");
```
pamapa, line 47: new Logger("State").create("fromStorageString");

src/State.ts
```diff
@@ -44,15 +44,15 @@ export class State {
...
         return new State(JSON.parse(storageString));
     }
 
     public static async clearStaleState(storage: StateStore, age: number): Promise<void> {
         const cutoff = Timer.getEpochTime() - age;
```
pamapa, line 52: const logger =  new Logger("State").create("clearStaleState");
...

src/User.ts
```diff
@@ -169,7 +169,7 @@ export class User {
     }
 
     public static fromStorageString(storageString: string): User {
-        Logger.debug("User", "fromStorageString");
+        Logger.debug("User.fromStorageString");
```
pamapa, line 172: new Logger("User").create("fromStorageString");

src/UserManager.ts
```diff
@@ -206,41 +207,37 @@ export class UserManager {
      * Returns promise to notify the opening window of response from the authorization endpoint.
      */
     public async signinPopupCallback(url = window.location.href, keepOpen = false): Promise<void> {
-        try {
-            await this._popupNavigator.callback(url, keepOpen);
-            this._logger.info("signinPopupCallback: successful");
-        }
-        catch (err) {
```
pamapa, line 213: why is no longer a catch needed?

src/UserManager.ts
```diff
@@ -317,14 +315,14 @@ export class UserManager {
      * Returns promise to query OP for user's current signin status. Returns object with session_state and subject identifier.
      */
     public async querySessionStatus(args: QuerySessionStatusArgs = {}): Promise<SessionStatus | null> {
+        const logger =this._logger.create("querySessionStatus");
```
pamapa, line 318: space `logger =this` -> `logger = this`

src/UserManagerEvents.ts
```diff
@@ -49,14 +49,12 @@ export class UserManagerEvents extends AccessTokenEvents {
     }
 
     public load(user: User, raiseEvent=true): void {
-        this._logger.debug("load");
```
pamapa, line 52: why removing all that debug log? Generally its better to log more than less...

src/utils/Logger.ts
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+/**
+ * Native interface
+ *
+ * @public
+ */
+export interface LoggerImplementation {
```
pamapa, line 9: i do not like the name `LoggerImplementation` its confusing as it is not an implementation, but an interface that a logger implementation must fullfill...

src/navigators/PopupWindow.ts
```diff
@@ -53,10 +53,8 @@ export class PopupWindow extends AbstractChildWindow {
     }
 
     public close(): void {
-        if (this._window) {
-            if (!this._window.closed) {
-                this._window.close();
-            }
+        if (this._window && !this._window.closed) {
+            this._window.close();
             this._abort.raise(new Error("Popup closed"));
```
pamapa, line 58: more than a log refactor, makes sense to have for that a separate commit...

src/navigators/IFrameWindow.ts
```diff
@@ -55,8 +55,8 @@ export class IFrameWindow extends AbstractChildWindow {
         if (this._frame) {
             if (this._frame.parentNode) {
                 this._frame.parentNode.removeChild(this._frame);
+                this._abort.raise(new Error("IFrame removed from DOM"));
             }
-            this._abort.raise(new Error("IFrame removed from DOM"));
```
pamapa, line 59: no log refactor, makes sense to have for that a separate commit...

src/AccessTokenEvents.ts
```diff
@@ -36,17 +37,17 @@ export class AccessTokenEvents {
                     expiring = 1;
                 }
 
-                this._logger.debug("load: registering expiring timer in:", expiring);
+                logger.debug("registering expiration timer");
```
pamapa, line 40: why not logging `expiring`?

src/AccessTokenEvents.ts
```diff
@@ -36,17 +37,17 @@ export class AccessTokenEvents {
...
             else {
-                this._logger.debug("load: canceling existing expiring timer because we're past expiration.");
+                logger.debug("canceling existing expiring timer because we're past expiration.");
                 this._expiringTimer.cancel();
             }
 
             // if it's negative, it will still fire
             const expired = duration + 1;
-            this._logger.debug("load: registering expired timer in:", expired);
+            logger.debug("registering expired timer");
```
pamapa, line 50: why not logging `expired`?

src/JsonService.ts
```diff
@@ -28,28 +28,29 @@ export class JsonService {
...
         try {
-            this._logger.debug("getJson, url:", url);
+            logger.debug("url:", url);
             response = await fetch(url, { method: "GET", headers });
         }
         catch (err) {
-            this._logger.error("getJson: network error");
-            throw new Error("Network Error");
+            logger.error("Network Error");
+            throw err;
```
pamapa, line 47: is err always from type `Error`?

src/MetadataService.ts
```diff
@@ -105,39 +104,38 @@ export class MetadataService {
     }
 
     protected async _getMetadataProperty(name: keyof OidcMetadata, optional=false): Promise<string | boolean | string[] | undefined> {
-        this._logger.debug("getMetadataProperty for: " + name);
+        const logger = this._logger.create(`getMetadataProperty('${name}')`);
```
pamapa, line 107: undercore in name?
->
```
const logger = this._logger.create(`_getMetadataProperty('${name}')`);
```

src/OidcClient.ts
```diff
@@ -123,42 +123,41 @@ export class OidcClient {
...
 
     public async readSigninResponseState(url: string, removeState = false): Promise<{ state: SigninState; response: SigninResponse }> {
-        this._logger.debug("readSigninResponseState");
+        const logger = this._logger.create("readSigninResponseState");
 
         const response = new SigninResponse(UrlUtils.readParams(url, this.settings.response_mode));
         if (!response.state) {
-            this._logger.error("readSigninResponseState: No state in response");
-            throw new Error("No state in response");
+            // https://github.com/microsoft/TypeScript/issues/46972
```
pamapa, line 130: this issue does not explain for me that code flow

src/SessionMonitor.ts
```diff
@@ -17,22 +17,17 @@ export class SessionMonitor {
     private _checkSessionIFrame?: CheckSessionIFrame;
 
     public constructor(private readonly _userManager: UserManager) {
-        if (!_userManager) {
```
pamapa, line 20: no logger refactor, should be in a separate commit

src/SessionMonitor.ts
```diff
@@ -153,36 +150,33 @@ export class SessionMonitor {
...
                 }
             }
             else {
-                this._logger.debug("_callback: Subject no longer signed into OP");
+                logger.debug("subject no longer signed into OP");
             }
 
             if (raiseEvent) {
                 if (this._sub) {
-                    this._logger.debug("_callback: SessionMonitor._callback; raising signed out event");
```
pamapa, line 173: why no more logging, generally we should log more than less...

pamapa:
Nice work! I really like the new `create` and `throw` functionally, they reduce copy/paste and make the code more readable...

src/utils/Logger.ts
```diff
@@ -0,0 +1,132 @@
...
+        return methodLogger;
+    }
+
+    private static _format(name: string, method?: string) {
+        const prefix = `[${name}]`;
+        return method ? prefix + ` ${method}:` : prefix;
+    }
+
+    // helpers for static class methods
+    public static debug(name: string, ...args: unknown[]): void {
```
pamapa, line 110: Are there still many places where these static log functions are used? Maybe its easier to replace them like `new Logger(xx).create(yy)`?

pamapa:
@kherock I will be some days offline, wish you a Merry Christmas...

src/User.ts
kherock: I wanted to distinguish static methods from instance methods, so I used the whole property accessor as the name in brackets in these cases.

src/UserManager.ts
kherock: parity with `signoutPopupCallback`'s behavior which also matches the `*RedirectCallback` and `*SilentCallback` methods. Catching the error here makes it difficult for the popup window to detect an error since the function is `void` return.

src/JsonService.ts
kherock: Yeah everything from the DOM should extend Error. Likewise I'm also pretty sure it's also always safe to cast caught errors as `Error` within this project as long as we trust crypto-js and jwt-decode to throw those which I believe they do

src/OidcClient.ts
kherock: yeah it was happening in too many places to include the comment each time, though I may refactor this to a `throw null` on the following line since this workaround feels too hacky

src/SessionMonitor.ts
kherock: I also didn't realize that we export SessionMonitor from the entrypoint so this check should stay anyway

src/SessionMonitor.ts
kherock: I'll add an else block logging that it's explicitly not raising an event, otherwise these lines weren't adding any useful info since raising the event already prints out everything you need to know, e.g.


```
[SessionMonitor] _callback: different subject signed into OP
[Event("User signed out"): raise
```

```
[SessionMonitor] _callback: same sub still logged in at OP, restarting check session iframe
                       (no further logs from this method)
```

src/AccessTokenEvents.ts
kherock: the log needed better phrasing and it felt simpler to just remove the value since it's already logged in the `init()` method immediately following this. I can add it back with some better phrasing

src/utils/Logger.ts
kherock: Not internally, so yeah that's probably a good idea.

pamapa:
great work, is its ready for merging?

pamapa (approved):

PR #320, opened 2022-01-12
POC: retry silent renew

### LINKED ISSUE #251: Waking up from sleep
`oidc-client-ts` does not seem to handle the computer going to sleep.  When the computer wakes up, typically the error that occurs is:

> [SilentRenewService] _tokenExpiring: Error from signinSilent: IFrame timed out without a response

and `oidc-client-ts` stops trying to silently refresh the token.

My main question is, should this be handled in my application or by `oidc-client-ts`?

If I need to handle in my application, first I would need to detect this condition. My thinking is to use the Page Visibility API.  Whenever the page becomes visible, I would check for a timeout problem in `oidc-client-ts`.  If this is detected, I would try to get `oidc-client-ts` to continue its silent refreshing with `UserManager.signinSilent()`.

It is possible (though unlikely) that the refresh token has timed out if the computer has been asleep for a long time. So if `UserManager.signinSilent()` fails, I would try again with `UserManager.signinRedirect()` however that would effectively restart my application.

Any ideas on how this can be handled?
If my application has to handle this, how could it detect that `oidc-client-ts` has timed out and is no longer doing silent refreshes?

Thanks for your help.

pamapa:
- retry silent renew

Closes/fixes #251

### Checklist


pamapa:
Missing: must ensure all code paths return an ErrorResponse in case the auth server returns not 200...
- `JsonService` throws ErrorResponse and Error (special cases)
- `IFrameWindow` seems not to throw ErrorResponse

src/SilentRenewService.ts
```diff
@@ -44,7 +49,15 @@ export class SilentRenewService {
         try {
             await this._userManager.signinSilent();
             logger.debug("silent token renewal successful");
-        } catch (err) {
+        }
+        catch (err) {
+            if (this._userManager.settings.retrySilentRenew && !(err instanceof ErrorResponse)) {
```
longsleep, line 54: I think it's not OK to always retry - some errors must stop the silent retrying - for example if consent is required or if not signed in at all.

longsleep (changes requested):

src/SilentRenewService.ts
kherock: If the authorization server is behaving correctly, this won't retry since the error would an `ErrorResponse` instance with `err.error === 'consent_required'` or `err.error === 'login_required'`, etc.

kherock:
I still haven't had time to really think through how to fix this issue, but I do feel like the proper solution shouldn't need a new configuration property.

Is it possible to give timeout errors their own error class or error code so that we can distinguish them from more severe errors (e.g. ID token mismatch or an OAuth error like invalid_grant)? I feel like it's always acceptable to retry when the server times out.

Also, I think this PR should also probably handle the case where a refresh token request fails due to the refresh token expiring and fall back to renewing with a prompt=none request.

pamapa:
I think it should be possible to split the different Errors into different stable API error classes. That would give the users of that library more insights.

I am in favor of moving the retry silent renew mechanisms into the library, as not everybody wants to have that we need a new configuration property. Alternatively a lot of people need to implement this by their own in probably different ways.

pamapa:
I have adapted the code to use a new ErrorTimeout, the configuration property is still part, i am in favor to drop it now in case we all agree...

Are there some other places we need to use the ErrorTimeout?

pamapa:
In case we want to distinguish a fetch promise error (timeout, other), we would need to extend it with special code to handle the timeout case. The fetch timeout seems also to differ from browser to browser (said defaults: firefox 60s, chrome 300s). And for the silentRequestTimeoutInSeconds we use a default of 10s, hmmm....

pamapa:
iframe part has been moved into MR #331

pamapa:
@kherock What is your opinion with my approach to handle silent renew with refresh token, aka. `fetchWithTimeout`? I will make a clean/new MR if we agree to go that direction...

pamapa:
retry silent renew for fetch timeout has been moved into MR #338

PR #321, opened 2022-01-13
Small 2

### LINKED ISSUE #314: signingKeys appears to be unused?
Hi,

Apologies if this turns out to be a non-issue but I was setting up a project locally and noticed that everything seems to work fine if I provide `metadata` as an option, but omit `signingKeys`. The `.well-known/openid-configuration` endpoint on the IDP that I'm using doesn't support CORS so they can't be coming from the jwks endpoint either.

I've had a look around the oidc-client-ts codebase and it looks like the `getSigningKeys` method on `MetadataService` isn't used, along with `resetSigningKeys`.

Am I correct with the above? And if so, should they be used?

oidc-client-js looks like it uses those methods - [OidcClient.processSigninResponse](https://github.com/IdentityModel/oidc-client-js/blob/dev/src/OidcClient.js#L129) calls [ResponseValidator.validateSigninResponse](https://github.com/IdentityModel/oidc-client-js/blob/dev/src/ResponseValidator.js#L32) which [leads to ](https://github.com/IdentityModel/oidc-client-js/blob/dev/src/ResponseValidator.js#L373) the response being validated from the keys retrieved from the JWKS endpoint, or the ones provided via config.

Thanks

pamapa:
- improve one log message
- docu fixes

Closes/fixes #314

### Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -49,9 +49,6 @@ provider:
   - authorization_endpoint
   - userinfo_endpoint
   - end_session_endpoint
-  - jwks_uri
-- [signingKeys](interfaces/UserManagerSettings.html#signingKeys) (which is the
-  `keys` property of the `jwks_uri` endpoint)
 - [metadataSeed](interfaces/UserManagerSettings.html#metadataSeed) can be used
   to seed or add additional values to the results of the discovery request.
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -17,6 +17,8 @@ removed.
     their behavior
 - restricted `response_type` to `code` flow only (PKCE remains optional)
   - as in oidc-client 1.x, OAuth 2.0 hybrid flows are not supported
+  - the property `signingKeys` is unused, unless the MetaDataService with this feature is used
+    outside of this library.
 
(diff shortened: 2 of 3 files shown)


src/utils/Timer.ts
```diff
@@ -23,7 +23,7 @@ export class Timer extends Event<[void]> {
         const expiration = Timer.getEpochTime() + durationInSeconds;
         if (this.expiration === expiration && this._timerHandle) {
             // no need to reinitialize to same expiration, so bail out
-            logger.debug("skipping since already initialized for expiration at", this.expiration);
```
kherock, line 26: This log is misleading since the check isn't necessarily enforcing that a duration value was already used once. e.g. this condition prevents a timer with duration with 10 seconds initialized 3 seconds ago from bing reinitialized when a new duration of 7 seconds is passed.

Now that I look at it, this whole early return seems unnecessary. I'm not sure what consequence it's avoiding.

kherock: Yeah I think the old message reads fine. perhaps it could be formatted as a date string but it's not really important to me

kherock (approved):

PR #328, opened 2022-01-14
Attempt to fix iframe leak

### LINKED ISSUE #138: heap memory grows over iterations of silent signin
if i repeatedly do silent sign-in, for example un-authenticated (result in login_required) (keep pressing renew button in angular app) ... then the **heap grows** and the source keeps adding silent-callback.html instances ...

see screen shots 

after investigation, it seems that if i make a change to I**FrameWindow.navigate** to **appendChild AFTER set the iframe.src**, then the GC can reclaim this heap and the silent-callback.html instances are removed ... however i can not explain why ... looking for help to 
understand what is happening here 

this **heap leak** needs to be fixed especially for spa sites

i made a fork with changes jeffschulzusc / oidc-client-ts

Screen Shot 2021-10-15 at 9 03 02 AM

Screen Shot 2021-10-15 at 9 02 44 AM

kherock:
Closes/fixes #138

I'm not sure if this fixes the whole problem but I'm also not sure what else will. This navigates the iframe to `about:blank` and waits for the page to load before removing it from DOM. In my testing, this allows for sources to unload from the "Sources" devtools panel but I'm not sure how to analyze the heap snapshots.

### Checklist

### DIFF
--- a/src/navigators/IFrameWindow.ts
+++ b/src/navigators/IFrameWindow.ts
@@ -54,8 +54,12 @@ export class IFrameWindow extends AbstractChildWindow {
     close(): void {
         if (this._frame) {
             if (this._frame.parentNode) {
-                this._frame.parentNode.removeChild(this._frame);
-                this._abort.raise(new Error("IFrame removed from DOM"));
+                this._frame.addEventListener("load", (ev) => {
+                    const frame = ev.target as HTMLIFrameElement;
+                    frame.parentNode?.removeChild(frame);
+                    this._abort.raise(new Error("IFrame removed from DOM"));
+                }, true);
+                this._frame.contentWindow?.location.replace("about:blank");
             }
             this._frame = null;
         }


pamapa:
I am looking at an other [implementation](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0e04ce3b0f93f001900d1056fc22bc18658a1bcb/lib/msal-browser/src/interaction_handler/SilentHandler.ts), maybe we can use some ideas.

1. createHiddenIframe + loadFrameSync + removeHiddenIframe as separate functions makes the code nice readable
2. authFrame.setAttribute("sandbox", "allow-scripts allow-same-origin allow-forms"); improves security
3. there seems to be no special handling about when to set `src` or change of sequence to `appendChild`
4. removing the iframe only with this condition seems reasonable `if (document.body === iframe.parentNode) {`
5. they append it differently `document.getElementsByTagName("body")[0].appendChild(authFrame);` Why?

What i do not understand in our code is why we throw an error when the iframe is removed in `close`
[this._abort.raise(new Error("IFrame removed from DOM"));](https://github.com/authts/oidc-client-ts/blob/74ee0bc2fd2e9b1d7f6bb935a33a1d65763f23ea/src/navigators/IFrameWindow.ts#L58)

pamapa:
The trick in the other [implementation](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/0e04ce3b0f93f001900d1056fc22bc18658a1bcb/lib/msal-browser/src/interaction_handler/SilentHandler.ts) seems to be the `monitorIframeForHash` function, it does not immediately remove the iframe when done, but after a certain time.

pamapa:
I am also not finding a better way than yours, without refactoring a lot. I am using now also you trick with the "Sources" devtools, to see if the iframe is gone. Otherwise it is possible to make a snapshot of the heap via "Memory" devtools. Should not grow each time a silent renew was done. I am using this trick in the settings to produce every 30s a silent renew:
```
{
// ...
// <MY_ACCESS_TOKEN_EXPIRE_TIME> is the total expire time of the access token, depends on the configuration in the auth server...
accessTokenExpiringNotificationTimeInSeconds: <MY_ACCESS_TOKEN_EXPIRE_TIME> - 30, // renew every 30s
}
```

pamapa:
Can you explain me why we need to call [this._abort.raise(new Error("IFrame removed from DOM"));](https://github.com/authts/oidc-client-ts/blob/74ee0bc2fd2e9b1d7f6bb935a33a1d65763f23ea/src/navigators/IFrameWindow.ts#L58)? When i remove it its seems to work too...

kherock:
Under typical usage, I don't think that event is capable of resulting in a thrown error, but it's good to have if we eventually need a mechanism to avoid leaking an unresolved promise created by `.navigate()` using the `close()` method. We currently only call `.close()` in two situations

1. In UserManager when `.navigate()` fails: https://github.com/authts/oidc-client-ts/blob/b8fa4439acbf4af37840284b3eaecccf583f8b35/src/UserManager.ts#L391-L392
2. After navigation success when `keepOpen` is false: https://github.com/authts/oidc-client-ts/blob/1d8254343748c367ef46e0992c510759ece990a2/src/navigators/AbstractChildWindow.ts#L71-L72

This test for the popup and redirect navigators is the should also throw the "removed from DOM" error if you were to run it against the iframe navigator instead.

https://github.com/authts/oidc-client-ts/blob/91ec29a2a306c4bad40179eccab962db329dd8d4/src/navigators/RedirectNavigator.test.ts#L27-L34

Badisi:
@kherock,

- In **situation 1**, the navigation could be stopped programmatically with the window being closed by let say an ads blocker, but it could also be stopped **manually** with the user closing the window himself. So should we call it an **error** in that case, as it was intentional ?

- In **situation 2.**, why would we throw an error in such a case ? because as you said : the navigation `succeeded`.

kherock:
Hi @Badisi - my comment was meant to highlight how neither of the two locations where we programmatically close the window will end up throwing an error when it's raised.

Keep in mind that with my second bullet, we're not throwing an error when we raise the event, we're only informing any unresolved promises that they _should_ throw if they're still pending. It's designed this way because the side effects of closing a successful window are similar enough to closing a window prematurely: we need to ensure that the window is closed/unloaded, event listeners are unregistered, and unresolved promises are rejected. We could refine these two flows separately but I don't think it's worth the complexity.

> but it could also be stopped manually with the user closing the window himself. So should we call it an error in that case, as it was intentional ?

As for handling a window closed by the user (or popup-blocking settings that the user is responsible for) - when an application requests to open a popup window, we don't know what the user's intent is, we only know that the application wants to obtain a sign in response. My stance is that anything which interrupts this process should be considered exceptional, and that includes the user's own actions. Also if the user's intent is to link their social account for example, closing the window would be _intentionally_ failing to accomplish that intent. They might have forgotten a password, the page might have been taking too long to load, etc.

Badisi:
> Keep in mind that with my second bullet, we're not throwing an error when we raise the event, we're only informing any unresolved promises that they should throw if they're still pending

My bad, I should have looked better at what `this._abort.raise()` was really doing 🙂

I was seeing errors in the console that looked like they were thrown and I thought it was due to that piece of code.

Screen Shot 2022-01-18 at 23 37 04

But looking at it now, I understand that the issue is instead related to the logger which prints the raised event to the console here : [Event.ts#L34](https://github.com/authts/oidc-client-ts/blob/1d8254343748c367ef46e0992c510759ece990a2/src/utils/Event.ts#L34).

In the case of an event being an **error**, it will be printed to the console with a call stack as if it was thrown.

pamapa (approved):

pamapa:
nice work

Badisi:
Issue was resolved and closed, but what do you guys think about my last [comment](https://github.com/authts/oidc-client-ts/pull/328#issuecomment-1015899998) ?
The logger should only log a message's error instead of the whole stack trace, don't you think ?

kherock:
For the `raise()` function, the logger should log the parameters that it's emitting, and unfortunately devtool consoles will format error objects with a stack trace most of the time. I do really think this is fine, especially since the debug/verbose logs aren't actually designed to be readable for end-users. I don't see much use for them beyond providing context for bug reports here.

If changing the log statement to log an array of arguments (which I believe devtools collapses by default) instead spreading them is more readable, I think that would be fine.

PR #329, opened 2022-01-14
fix(RedirectNavigator): wait until the window is unloading to resolve

kherock:
### Checklist

While testing the new `activeNavigator` prop in react-oidc-context, I noticed that `signinRedirect` returns synchronously and was causing the component to immediately rerender since the redirect navigator would close immediately after invocation. To fix this, I've made it so that `navigate()` returns a promise that only resolves as the page is exiting. The timing of `beforeunload` is late enough that the tasks in the async queue waiting for the promise to resolve are never run.

### DIFF
--- a/src/UserManager.test.ts
+++ b/src/UserManager.test.ts
@@ -36,22 +36,6 @@ describe("UserManager", () => {
                 revocation_endpoint: "http://sts/oidc/revoke",
             },
         });
-
-        const location = Object.defineProperties({}, {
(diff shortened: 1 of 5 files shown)


pamapa (approved):

pamapa:
nice work

nvmnghia:
Could you explain this PR a lil bit @kherock? Just want to know why `signoutRedirect()` must be async, when what it does is redirecting the website. Thx!

PR #330, opened 2022-01-16
fix(RedirectNavigator): wait for 'unload' instead of 'beforeunload'

kherock:
### Checklist

I didn't test the fix in #329 thoroughly enough, in Edge (and presumably Chrome) I was still seeing my React app rerender and trigger another redirect, canceling the unload, and causing an infinite loop. Ideally the promise should just never resolve by the time the page is unloading, but I was having a hard time writing tests using functions that never resolve.

The use of `'unload'` here is discouraged but this case in particular seems fine to me. We don't care about browser consistency since we don't even necessarily need the promise to ever resolve, and we are truly waiting for the page to unload rather than the page visibility changing.

### DIFF
--- a/src/navigators/RedirectNavigator.ts
+++ b/src/navigators/RedirectNavigator.ts
@@ -32,7 +32,7 @@ export class RedirectNavigator implements INavigator {
                 this._logger.create("navigate");
                 const promise = new Promise((resolve, reject) => {
                     abort = reject;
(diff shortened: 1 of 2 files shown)


pamapa (approved):

pamapa:
good catch, i will make soon another rc release so we can test that too...

PR #331, opened 2022-01-17
Retry silent renew 1 iframe timeout

pamapa:
This is the first part of two MR. It implements the case to retry silent renew process in case of iframe timeouts. The case with the network timeout (via refresh token) will be done in the second MR.
I have tested it multiple hours with reduce silent renew time (every 30s). Turning off network via inspector knob and also physically disconnected the network. Looks really nice.
As suggested previously, i have **not** added any new property in the settings.

Closes/fixes part of #251

### Checklist


kherock (approved):
LGTM

pamapa:
As this might try every 5s a silent renew, we must ensure we do not leak anything. Thus #328 needs to merged first...

PR #338, opened 2022-01-20
retry silent renew for fetch timeout

pamapa:
By default a `fetch()` request timeouts at the time indicated by the browser. In Chrome a network request timeouts at 300 seconds, while in Firefox at 90 seconds.
The implementation is inspired by this [blog](https://dmitripavlutin.com/timeout-fetch-request/).

Closes/fixes part of #251

### Checklist

### DIFF
--- a/src/JsonService.ts
+++ b/src/JsonService.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-import { ErrorResponse } from "./errors";
+import { ErrorResponse, ErrorTimeout } from "./errors";
 import { Logger } from "./utils";
 
 /**
@@ -27,6 +27,33 @@ export class JsonService {
         }
     }
 
+    protected async fetchWithTimeout(input: RequestInfo, init: RequestInit & { timeoutInSeconds?: number } = {}) {
+        const { timeoutInSeconds, ...initFetch } = init;
+        if (!timeoutInSeconds) {
+            return await fetch(input, initFetch);
+        }
+
+        const controller = new AbortController();
+        const timeoutId = setTimeout(() => controller.abort(), timeoutInSeconds * 1000);
+
+        try {
+            const response = await fetch(input, {
+                ...init,
+                signal: controller.signal,
+            });
+            return response;
+        }
+        catch (err) {
+            if (err instanceof DOMException && err.name === "AbortError") {
+                throw new ErrorTimeout("Network timed out");
+            }
+            throw err;
+        }
+        finally {
+            clearTimeout(timeoutId);
+        }
+    }
+
     public async getJson(url: string, token?: string): Promise<Record<string, unknown>> {
         const logger = this._logger.create("getJson");
         const headers: HeadersInit = {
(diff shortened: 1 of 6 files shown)


pamapa:
The __8 seconds__ for the default request timeout is related to the SilentRequestTimeoutInSeconds, which defaults to 10 seconds.
Maybe its better to **only** use that default for the silent renew request and leave the other requests (code flow, metadata fetch) as is...

pamapa:
The refactor of the argument to object of useRefreshToken has been moved into this MR #340. Which must be merged first, then i will rebase this one...

src/UserManager.ts
```diff
@@ -263,7 +263,11 @@ export class UserManager {
     }
 
     protected async _useRefreshToken(state: RefreshState): Promise<User> {
-        const response = await this._client.useRefreshToken(state);
+        const response = await this._client.useRefreshToken({
+            state,
+            timeoutInSeconds: this.settings.silentRequestTimeoutInSeconds
+                ? Math.max(this.settings.silentRequestTimeoutInSeconds - 2, 1) : 8,
```
kherock, line 269: Why is this set to subtract 2 seconds from `silentRequestTimeoutInSeconds`?

pamapa: You are right makes no sense, in this scenario it is actually `silentRequestTimeoutInSeconds`. I will fix that when i do the rebase.

pamapa:
Lets merge #339 first, such that i do not have to hardcode the 10s default...

longsleep (approved):
LGTM 👍

src/UserManager.ts
```diff
@@ -263,7 +263,11 @@ export class UserManager {
     }
 
     protected async _useRefreshToken(state: RefreshState): Promise<User> {
-        const response = await this._client.useRefreshToken({ state });
+        const response = await this._client.useRefreshToken({
+            state,
+            timeoutInSeconds: this.settings.silentRequestTimeoutInSeconds
+                ? this.settings.silentRequestTimeoutInSeconds : 10,
```
kherock, line 269: With #339 `this.settings.silentRequestTimeoutInSeconds` should always have a defined value and we can avoid hardcoding this `10` value.

src/JsonService.ts
```diff
@@ -74,7 +98,7 @@ export class JsonService {
         return json;
     }
 
-    public async postForm(url: string, body: URLSearchParams, basicAuth?: string): Promise<Record<string, unknown>> {
+    public async postForm(url: string, body: URLSearchParams, basicAuth?: string, timeoutInSeconds?: number): Promise<Record<string, unknown>> {
```
kherock, line 101: This class is internal so we can always come back to this later, but just want to criticize that this function now takes in 4 parameters 😬

src/JsonService.ts
```diff
@@ -27,6 +27,30 @@ export class JsonService {
...
+        }
+
+        try {
+            const controller = new AbortController();
+            const timeoutId = setTimeout(() => controller.abort(), timeoutInSeconds * 1000);
+            const response = await fetch(input, {
+                ...init,
+                signal: controller.signal,
+            });
+            clearTimeout(timeoutId);
```
kherock, line 43: I think this line would be better in a `finally` block. Lines `37` and `38` can be moved outside of the `try` since they aren't expected to throw.

src/JsonService.ts
```diff
@@ -27,6 +27,30 @@ export class JsonService {
...
+            return response;
+        }
+        catch (err) {
+            if ((err as Error).name === "AbortError") {
```
kherock, line 47: I think this is fine, though you can make this a tad stricter with `if (err instanceof DOMException && err.name === 'AbortError')`.

src/JsonService.ts
pamapa: done in separate MR, which can be merged after this one. See #349

kherock (approved):

PR #339, opened 2022-01-20
refactor: consolidate defaults into settings

pamapa:
When working on timeout for fetch i saw that it would be better to have access to the `silentRequestTimeoutInSeconds` including default applied. This was something on my to-do list for some time. And it is easier to verify the defaults, as you only have to look into the settings and not search all over the code...

### Checklist

### DIFF
--- a/src/OidcClientSettings.ts
+++ b/src/OidcClientSettings.ts
@@ -111,7 +111,7 @@ export interface OidcClientSettings {
 export class OidcClientSettingsStore {
     // metadata
     public readonly authority: string;
-    public readonly metadataUrl: string | undefined;
+    public readonly metadataUrl: string;
     public readonly metadata: Partial<OidcMetadata> | undefined;
     public readonly metadataSeed: Partial<OidcMetadata> | undefined;
     public readonly signingKeys: SigningKey[] | undefined;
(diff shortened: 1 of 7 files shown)


src/OidcClientSettings.ts
```diff
@@ -12,6 +12,7 @@ const DefaultClientAuthentication = "client_secret_post";
 const DefaultResponseMode = "query";
 const DefaultStaleStateAgeInSeconds = 60 * 15;
 const DefaultClockSkewInSeconds = 60 * 5;
+const DefaultOidcMetadataUrlPath = ".well-known/openid-configuration";
```
kherock, line 15: This isn't a default value, it's part of the protocol. I think you could actually just get rid of this variable and inline it in the constructor below.

src/UserManagerSettings.ts
```diff
@@ -101,12 +108,12 @@ export class UserManagerSettingsStore extends OidcClientSettingsStore {
         const {
             popup_redirect_uri,
```
kherock, line 109: You could simplify some of these further with
```suggestion
            popup_redirect_uri = args.redirect_uri,
```

PR #341, opened 2022-01-21
Improve iframe creation

pamapa:
Saw this when @kherock was fixing the iframe leak. This makes the code more readable and the sandbox improves security and is best practice.

### Checklist

### DIFF
--- a/src/navigators/IFrameWindow.ts
+++ b/src/navigators/IFrameWindow.ts
@@ -43,6 +43,7 @@ export class IFrameWindow extends AbstractChildWindow {
         iframe.style.top = "0";
         iframe.width = "0";
         iframe.height = "0";
+        iframe.setAttribute("sandbox", "allow-scripts allow-same-origin allow-forms");
 
         window.document.body.appendChild(iframe);
         return iframe;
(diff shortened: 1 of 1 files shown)


src/navigators/IFrameWindow.ts
```diff
@@ -29,19 +29,24 @@ export class IFrameWindow extends AbstractChildWindow {
         super();
         this._timeoutInSeconds = silentRequestTimeoutInSeconds;
 
-        this._frame = window.document.createElement("iframe");
+        this._frame = IFrameWindow.createHiddenIframe();
         this._window = this._frame.contentWindow;
+    }
+
+    private static createHiddenIframe(): HTMLIFrameElement {
```
kherock, line 36: Do you think it would be a good idea to make this public and re-use it in `CheckSessionIFrame`? I haven't been able to figure out a good way to reduce the code that is duplicated over there.

pamapa: Yes it might makes sense to re-use. I do not plan to touch `CheckSessionIFrame` any time soon. Lets merge this as is, we can pretty easily make this function later public, when there is an actual need...

PR #349, opened 2022-01-24
refactor: align arguments in json service

pamapa:
As suggested in MR #338. To be merged after #338...

### Checklist

### DIFF
--- a/src/JsonService.ts
+++ b/src/JsonService.ts
@@ -12,6 +12,22 @@ export type JwtHandler = (text: string) => Promise<Record<string, unknown>>;
 /**
  * @internal
  */
+export interface GetJsonOpts {
+    token?: string;
+}
+
+/**
+ * @internal
+ */
+export interface PostFormOpts {
+    body: URLSearchParams;
+    basicAuth?: string;
+    timeoutInSeconds?: number;
+}
+
+/**
+ * @internal
+ */
 export class JsonService {
     private readonly _logger = new Logger("JsonService");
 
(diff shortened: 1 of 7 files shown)


src/JsonService.ts
```diff
@@ -74,7 +122,12 @@ export class JsonService {
         return json;
     }
 
-    public async postForm(url: string, body: URLSearchParams, basicAuth?: string): Promise<Record<string, unknown>> {
+    public async postForm({
+        url,
```
kherock, line 126: I don't have a particularly strong opinion on the API here, but I think keeping `url: string` for the first arguments would be nice it that it mirrors `fetch` and would make many of the tests simpler to write.

src/TokenClient.test.ts
```diff
@@ -87,9 +87,9 @@ describe("TokenClient", () => {
             expect(generateBasicAuthSpy).toHaveBeenCalledWith("client_id", "client_secret");
             expect(getTokenEndpointMock).toBeCalledWith(false);
             expect(postFormMock).toBeCalledWith(
-                "http://sts/token_endpoint",
-                expect.any(URLSearchParams),
-                expect.stringContaining(""),
+                expect.objectContaining({
+                    url: "http://sts/token_endpoint",
+                }),
```
kherock, line 92: This test is a less strict than it was before, the previous revision had checks for URLSearchParams body and the basic auth string. There's several other places within this file of this occurring.

kherock (approved):
Nice, thank you

pamapa:
Thanks for reviewing, it is much better like this :-)

PR #365, opened 2022-02-03
fix: add scope to refresh tokens

### LINKED ISSUE #364: scope in refresh tokens
I'm using the `oidc-client-ts` with Azure AD, and I stumbled on an issue refresh tokens. 

From what I under from https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#refresh-the-access-token and https://docs.microsoft.com/en-us/answers/questions/192699/confusing-error-message-on-failed-refresh-token-re.html Azure AD requires the scope to be sent when refreshing the access token.

If change [OidcClient.ts#L168-L171](https://github.com/authts/oidc-client-ts/blob/main/src/OidcClient.ts#L168-L171) to include the scope as below this issue is resolved for me.

```ts
const result = await this._tokenClient.exchangeRefreshToken({
  refresh_token: state.refresh_token,
  scope: state.scope,
  timeoutInSeconds,
});
```

Would this be something that would make sense to add?

I'm new to OIDC, so please excuse any misunderstanding on my part.

Thanks for maintaining this library. It helped me quickly to get an OIDC flow implemented.

kasperp:
As described in #364 Azure AD requires the scope to be sent when refreshing the access token. This adds scope to the `exchangeRefreshToken` call from `useRefreshTokens`

Closes/fixes #364

### Checklist

### DIFF
--- a/src/OidcClient.test.ts
+++ b/src/OidcClient.test.ts
@@ -359,7 +359,9 @@ describe("OidcClient", () => {
             const tokenResponse = {
                 access_token: "new_access_token",
             };
-            jest.spyOn(subject["_tokenClient"], "exchangeRefreshToken").mockResolvedValue(tokenResponse);
+            const exchangeRefreshTokenMock =
+                jest.spyOn(subject["_tokenClient"], "exchangeRefreshToken")
+                    .mockResolvedValue(tokenResponse);
             jest.spyOn(JwtUtils, "decode").mockReturnValue({ sub: "sub" });
             const state = new RefreshState({
                 refresh_token: "refresh_token",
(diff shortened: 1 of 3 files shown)


src/OidcClient.test.ts
```diff
@@ -371,6 +373,10 @@ describe("OidcClient", () => {
             const response = await subject.useRefreshToken({ state });
 
             // assert
+            expect(exchangeRefreshTokenMock).toHaveBeenCalledWith( {
+                refresh_token: "refresh_token",
+                scope: "openid",
+            });
```
kasperp, line 379: added an assert to an existing unit test, let me know if you'd rather have this tested differently.

kherock (approved):
Hey, sorry for the long wait on this. Thanks for the PR, we'll try to release this soon.

PR #388, opened 2022-02-19
chore(package.json): add module field

Badisi:
This PR is simply adding a `module` field, to the `package.json` file, which points to the `esm` version of the library.

----

#### Reasoning

While trying to import `oidc-client-ts` (as a module) over a CDN I noticed the following:

```html
<script type="module" src="https://unpkg.com/oidc-client-ts@latest?module"></script>

=> Package oidc-client-ts@2.0.1 does not contain an ES module
``` 

The issue here is that the `package.json` file of `oidc-client-ts` is missing a `module` field.

This field is not official but is a common convention among bundlers to designate how to import an ESM version ([stackoverflow#47537198](https://stackoverflow.com/a/47537198)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "homepage": "https://github.com/authts/oidc-client-ts#readme",
   "license": "Apache-2.0",
   "main": "dist/umd/oidc-client-ts.js",
+  "module": "dist/esm/oidc-client-ts.js",
   "types": "dist/types/oidc-client-ts.d.ts",
   "exports": {
     ".": {


kherock:
This field is only preferred in webpack 4 and extremely old versions of rollup plugins. Every up to date bundler within the past few years should be able to resolve ESM via the `exports` map.

The main CJS export should function just fine for the situations where `exports` maps aren't supported. I don't think `"module"` is a necessary convention to adopt here in 2022.

PR #417, opened 2022-03-11
fix: handle case of refresh token response not containing id_token

hughns:
Currently an error occurs and token refresh fails if the token response doesn't contain the `id_token` yet the `scope` contains `openid`.

This is allowed by the OIDC spec (relevant link is in the code).


pamapa:
LGTM

kherock (approved):

pamapa:
great work, thanks for contributing!

PR #426, opened 2022-03-15
allow to get and process claims when scope does not contain openid

### LINKED ISSUE #415: Automatic silent renew deletes profile from stored state
Hello,

I noticed that the property `profile` of the object stored inside the session storage is overwritten when a silent renew is triggered. It looks like the response of the silent renew just overwrites the whole object.

State stored in session storage after signin with redirect callback:
image

State stored in session storage after first silent renew:
image

I also enabled the option `loadUserInfo` of the UserManager, but the silent renew doesn't seem to respect it.
It would be great if the stored profile would remain in place even after the silentRenew.

pamapa:
Closes/fixes #415

### Checklist

### DIFF
--- a/src/ResponseValidator.test.ts
+++ b/src/ResponseValidator.test.ts
@@ -310,7 +310,7 @@ describe("ResponseValidator", () => {
             });
         });
 
-        it("should not run if request was not openid", async () => {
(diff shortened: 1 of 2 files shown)


kherock (approved):
LGTM

PR #439, opened 2022-03-23
feature(nonce): allow nonce fowarding through signin

### LINKED ISSUE #438: support custom nonce in Authentication Request (authorization code flow)
I need to be able to specify the nonce present in my ID Tokens. With this, one of my services can perform an additional challenge/check with my users. I want to use _±public cryptographic key_ generated by [libsodium](https://www.npmjs.com/package/libsodium-wrappers) for my nonce.

My challenge is based on the fact that the nonce given during authorization request is forwarded as is inside the ID Token
(see [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) nonce argument) and thus protected by ID Token signature.

I think the following API change would be good to have (aka. supporting nonce in `ExtraSigninRequestArgs`):
```typescript
import { UserManager } 'oidc-client-ts';

async function requestIdToken() {
  const manager =  new UserManager({
      // [...]
    });

  const nonce: string = // generate nonce using libsodium + url safe unpadded base64
  await manager.signinRedirect({ nonce });
}

async function handleRedirection() {
  const manager =  new UserManager({
      // [...]
    });

  await manager.signinRedirectCallback();
  const idToken = (await manager.getUser()).id_token;
  const customNonce: string = // extract nonce from idToken
}
```

I already have a fork ready if this proposal is acceptable

JMounier:
fixes #438

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -54,6 +54,8 @@ export interface CreateSigninRequestArgs {
     // (undocumented)
     max_age?: number;
     // (undocumented)
+    nonce?: string;
+    // (undocumented)
     prompt?: string;
     // (undocumented)
     redirect_uri?: string;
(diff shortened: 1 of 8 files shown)


pamapa:
Why do the workflow jobs: build + test not automatically start? hmmm

kherock (approved):
@pamapa you have to approve runs for first-time contributors! There's a button at the bottom of the PR for it.

Thanks for the contribution @JMounier!

pamapa:
@kherock Thanks for triggering the CI. Hmm i thought previously there was a button in the bottom area, but no longer, anyway.

PR #446, opened 2022-03-27
Add extra sign out query params

felipereyel:
## Adds configuration for extra signout query params.

Currently this project only has extraQueryParams that are passed to both signin and signout requests but sometimes you just want to pass to one or another and this PR allows to pass query params only to signout.

### Example:
I want to logout of a keycloak server but keycloak does not recognize the post_logout_redirect_uri query param but recognizes the redirect_uri (implementation specific) so now someone can just pass this query param to the

### Links
- https://github.com/keycloak/keycloak/issues/10135
- https://github.com/keycloak/keycloak/discussions/9950 

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -115,7 +115,7 @@ export class ErrorTimeout extends Error {
 export type ExtraSigninRequestArgs = Pick<CreateSigninRequestArgs, "nonce" | "extraQueryParams" | "extraTokenParams" | "state">;
 
 // @public (undocumented)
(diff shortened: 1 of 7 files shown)


pamapa:
Thanks for contributing.

I do not think that patch is needed. As you can pass signout only extraParams already when calling `mgr.signoutRedirect(args: SignoutRedirectArgs = {})`. In your case: do not store it in the settings, add it as parameter when calling the signout out function.

felipereyel:
@pamapa thats it. Thank you for your help.

PR #469, opened 2022-04-06
Return of UserManager.createSigninRequest

josundt:
_[x] This PR does not change, only augments public APIs with one new method._

Today I attempted to migrate from `oidc-client` to `oidc-client-ts` in our LOB products, and I ended up with only _one unresolvable issue_.

It looks like the method `UserManager.createSigninRequest(args: SigninRequestArgs): Promise<SigninRequest>` was "lost in translation" somewhere after you forked `oidc-client` and converted to TypeScript, or you may have decided that it should be changed to an internal feature.

Our applications actually depend on this method since we sometimes provide a "login" hyperlink in the UI. 
We get the URL for the login hyperlink using `(await userManager.createSigninRequest({...})).url`.
This functionality is unfortunately not available anymore.
 
In this PR I have added the method back for feature parity with the legacy client, and to solve the issue in our apps.
The added method simply relays to `_client.createSigninRequest(args: SigninRequestArgs): Promise<SigninRequest>` .

PS! I have added unit tests that covers the new code.
Hoping for approval and a quick package patch release! 👍

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -858,6 +858,7 @@ export class UserManager {
     clearStaleState(): Promise<void>;
     // (undocumented)
     protected readonly _client: OidcClient;
+    createSigninRequest(args: CreateSigninRequestArgs): Promise<SigninRequest>;
     get events(): UserManagerEvents;
     // (undocumented)
     protected readonly _events: UserManagerEvents;
--- a/src/UserManager.test.ts
+++ b/src/UserManager.test.ts
@@ -11,6 +11,7 @@ import type { UserProfile } from "./User";
 import { WebStorageStateStore } from "./WebStorageStateStore";
 import type { SigninState } from "./SigninState";
 import type { State } from "./State";
+import type { SigninRequest, SigninRequestArgs } from "./SigninRequest";
 
 import { mocked } from "jest-mock";
 
@@ -155,6 +156,24 @@ describe("UserManager", () => {
             await expect(subject.revokeTokens())
                 // assert
                 .resolves.toBe(undefined);
+        });
+    });
+
+    describe("createSigninRequest", () => {
+        it("should relay to _client.createSigninRequest", async () => {
+            // arrange
+            const args = {} as SigninRequestArgs;
+            const resultMock = {} as SigninRequest;
+            const spy = jest.spyOn(subject["_client"], "createSigninRequest")
+                .mockResolvedValue(resultMock);
+
+            // act
+            const signinRequest = await subject.createSigninRequest(args);
+
+            // assert
+            expect(spy).toBeCalledTimes(1);
+            expect(spy).toBeCalledWith(args);
+            expect(signinRequest).toBe(resultMock);
         });
     });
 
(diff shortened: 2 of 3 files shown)


josundt:
Hello and thanks for the feedback.
I've done some more digging into our own application and found that I provided some false information earlier - which really undermined the whole rationale for my request.

We don't need authorize URLs generated for _**hyperlinks**_ after all - and I see your point with the state param storage side-effect if it was used for hyperlinks as I first believed.

Our application actually needs to generate authorize URLs to integrate with our SPA application's _**router**_ to perform a _redirect_. The SPA framework being used allows a route definition to execute a (sync or async) function with a given signature. For the route execution to result in a _"browser redirect"_, the object (or Promise of object) returned from this function needs to contain a specific property with the redirect URL.

For this purpose the `UserManager.createSigninRequest(args: SigninRequestArgs): Promise<SigninRequest>` - which was a public method in the legacy oidc-client's `UserManager` class - did exactly what is required for our application to integrate the SPA router with a redirect to login (authorize URL).

I believe I have shown that this somewhat lower-level method is useful in certain scenarios like ours, and therefore I hope you can add it back - since it was a part of the legacy `UserManager` that was actually being used by consumers.
I believe my PR adds this back the way it worked in `oidc-client`.

FYI: This is what I have done as a temporary workaround in our application for now:
```typescript
(await ((this.userManager as Record<string, any>)["_client"] as OidcClient).createSigninRequest(signinRequestOptions)).url
```
This is of course very ugly.

I hope I provided some better reasoning for accepting the PR this time.

pamapa:
I am still not convinced and do not understand the problem. What are you doing with the result of `createSigninRequest`? and why don't you operate on a instance of `OdcClient`?

josundt:
This was an especially useful integration point for other 3rd party frameworks in the legacy `UserManager`.

When integrating `oidc-client` with other 3rd party components like SPA frameworks, navigation orchestration (routing, browser history) is an integral part of that other framework's API. 

When integrating `oidc-client-ts` with search frameworks, it is sometimes desirable to separate duties.
* `oidc-client-ts`: responsible for the URL generation
* the SPA framework: responsible for executing the actual browser navigation through it's integral navigation APIs.

`UserManager`'s `signinRedirect`, `signinPopup`, `signinSilent` methods both generate URLs **AND** trigger the browser navigation, and thereby is not suitable for integrating properly with the navigation APIs of the 3rd party framework.

In the legacy package I never had to relate to `OidcClient`, `UserManager` was the only class I really had to be concerned about. (I guess `OidcClient` was a base class for `UserManager` in the legacy package.)

I still have a hard time understanding why you insist on removing something that was useful and being used in the legacy `UserManager`. Maybe I could accomplish the same using the `OidcClient` but it seems like this would be a more complicated solution.

I hope I have been able to prove the value of this.

josundt:
OK, as long as there's a new solution that suffice, I guess I'm happy.
I have tested that it works - so this PR can be abandoned.

From the discussion in this PR, I still hope to have proven the value of exposing some lower-level public API where _URL generation_ and _browser navigation_ are separate responsibilities, since it is very useful for integration with other frameworks. This would be a great improvement compared to the legacy package.

The methods of such an API could maybe be a bit less available than the standard methods on the `UserManager`,  but could probably be a bit more intuitive than requiring `new OidcClient(userManager.settings)`.

I hope these are thoughts that you can take into consideration in future design decisions for the package.

And I may have forgotten to mention: Thanks for taking on the work of keeping this library alive and improving it!
I really like what I've seen this far - especially the typings :-)

pamapa:
Thanks for the feedback and the nice words!

PR #470, opened 2022-04-06
fix(CheckSessionIFrame): normalize iframe origin before comparing to the MessageEvent origin

### LINKED ISSUE #465: Session change event is not raised
Case: 
Say, the ID provider is deployed on an origin id.com (IdentityServer4-based). 

In the SPA, I subscribe to a session changed event, but when signed out from IDP using Single Sign Out, the event is not raised. 

The "changed" message is delivered from the child iframe (check session) to the main window, but the event still is not raised. 

monitorSession = true,
anonymousMonitorSession = true

This was working with oidc-client-js. 
What could be the problem? 

Thanks.

kherock:
Closes/fixes #465

### Checklist

This normalizes the stored iframe origin with `new URL` which fixes the comparison to the origin in the `MessageEvent` handler later.

The existing `CheckSessionIFrame` implementation is still a bit messy. I'd like to clean up its API a bit before adding unit tests for it.

### DIFF
--- a/src/CheckSessionIFrame.ts
+++ b/src/CheckSessionIFrame.ts
@@ -20,8 +20,8 @@ export class CheckSessionIFrame {
         private _intervalInSeconds: number,
         private _stopOnError: boolean,
     ) {
-        const idx = url.indexOf("/", url.indexOf("//") + 2);
-        this._frame_origin = url.substr(0, idx);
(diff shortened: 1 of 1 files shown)


pamapa (approved):

pamapa:
Thanks for fixing this!

PR #499, opened 2022-04-22
Remove the event listener for unloading when navigating

shakeelsoogun:
I got a somewhat similar issue to what was reported with #329 and #330. When calling `signinRedirect`, on Chrome everything seems fine. With Firefox (at least on v99.0.1), what I see happening is that the page does redirect to the OIDC provider, but oddly an infinite barrage of API requests get made when they shouldn't - as if its trying to reload my website over and over. I can only guess this must be a bug in Firefox itself to do with making an event listener before navigating or something?

I tried listening to different events (e.g. `beforeunload`, `visibilitychange`, etc. instead of `unload`), performing the window event listening outside of the promise function (to see if ensuring that its done before navigate could help) but neither of these make a difference. The only thing that seems to fix this is actually removing the event listening and making the promise never resolve. 

The tests here end up suffering however; as the author of #329 pointed out, its not easy to write proper tests for a promise that never resolves. The only thing I could do here was wait a "significant enough amount of time" that things should have executed (definitely not that nice), unless there's a better idea of how to carry this out?

Closes/fixes #issue

### Checklist

### DIFF
--- a/src/UserManager.test.ts
+++ b/src/UserManager.test.ts
@@ -161,7 +161,14 @@ describe("UserManager", () => {
     describe("signinRedirect", () => {
         it("should redirect the browser to the authorize url", async () => {
             // act
-            await subject.signinRedirect();
+            const spy = jest.fn();
+            subject.signinRedirect().finally(spy);
+
+            // signinRedirect is a promise that will never resolve (since we
+            // want it to hold until the page has redirected), so we use a
+            // timeout to wait for the majority of the code to have been run
+            // before proceeding with the rest of the test.
+            await new Promise(resolve => setTimeout(resolve, 200));
 
             // assert
             expect(window.location.assign).toHaveBeenCalledWith(
@@ -171,6 +178,9 @@ describe("UserManager", () => {
             const state = new URL(location).searchParams.get("state");
             const item = await userStoreMock.get(state!);
             expect(JSON.parse(item!)).toHaveProperty("request_type", "si:r");
+
+            // We check to make sure the promise has not resolved
+            expect(spy).not.toHaveBeenCalled();
         });
 
         it("should pass navigator params to navigator", async () => {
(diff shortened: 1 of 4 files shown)


pamapa:
@kherock Can you have a look at this?

src/UserManager.test.ts
```diff
@@ -161,7 +161,14 @@ describe("UserManager", () => {
...
             // act
-            await subject.signinRedirect();
+            const spy = jest.fn();
+            subject.signinRedirect().finally(spy);
+
+            // signinRedirect is a promise that will never resolve (since we
+            // want it to hold until the page has redirected), so we use a
+            // timeout to wait for the majority of the code to have been run
+            // before proceeding with the rest of the test.
+            await new Promise(resolve => setTimeout(resolve, 200));
```
kherock, line 171: Could you replace this with a promise waiting for the window to unload

```ts
await once(window, 'unload')
```

and update the mock implementation of the `unload` event to happen after a 200ms delay instead?

kherock (commented):
Hello and thank you @shakeelsoogun!

I think the testing approach is fine, though I think it might be cleaner to move the timeout into the mock `location.assign`/`location.replace` implementations we've set up.

src/UserManager.test.ts
kherock: Sorry, I probably should have suggested the import path too! 

```js
// https://nodejs.org/api/events.html#eventsonceemitter-name-options
import { once } from 'events';
```


It should be interoperable with browser `EventTarget` interfaces. If that doesn't work, what you have looks good 👍

shakeelsoogun: Ahh ok, didn't know about that! Much nicer looking than my version 👍🏾

kherock (approved):

pamapa:

PR #514, opened 2022-04-29
feat: add ability to specify iframe script origin and iframe notify parent for silent auth

pavliy:
**The problem**
Both - oidc-client-js & oidc-client-ts - support silent flow only within the same domain. However in my case - apps are hosted on different domains. It's possible to address it by specifying 2 new settings which will allow to overcome issue and make all things work even with different domains.

**What for**
I'm working on micro frontends implementation currently. 
- Architecture is based on webpack module federation and web components
- There is a shell hosted on on https://app-a.test.com
- And there is micro frontend app loaded through the shell, hosted as https://app-b.test.com
- Shell and each micro app - are separate relying parties which have their own client id, redirect urls etc.  Users can also access them directly.

The main goal is to avoid explicit login flow for getting micro app access token (user does not need to enter credentials once again)
Though, both apps are separate relying parties, token is issued by the same authority.

So the main idea:
- User logs in to shell: https://app-a.test.com and token is saved into session storage
- User clicks the link to load micro app - code is loaded using module federation, which includes oidc-client load dependency as well
- Silent auth flow happens to get token for micro app. And in result - it's also saved in session storage 

BTW, I'll add unit tests soon to cover updated scenarios
### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ node_modules/
 .DS_Store
 .vscode/
 temp/
+.history/
--- a/src/UserManagerSettings.ts
+++ b/src/UserManagerSettings.ts
@@ -36,6 +36,10 @@ export interface UserManagerSettings extends OidcClientSettings {
     /** The methods window.location method used to redirect (default: "assign") */
     redirectMethod?: "replace" | "assign";
 
+    /** The target to pass while calling postMessage inside iframe for callback */
+    iframeNotifyParentOrigin?: string;
+    iframeScriptOrigin?: string;
+
     /** The URL for the page containing the code handling the silent renew */
     silent_redirect_uri?: string;
     /** Number of seconds to wait for the silent renew to return before assuming it has failed or timed out (default: 10) */
@@ -86,6 +90,9 @@ export class UserManagerSettingsStore extends OidcClientSettingsStore {
     public readonly popupWindowTarget: string;
     public readonly redirectMethod: "replace" | "assign";
 
+    public readonly iframeNotifyParentOrigin: string | undefined;
+    public readonly iframeScriptOrigin: string | undefined;
+
     public readonly silent_redirect_uri: string;
     public readonly silentRequestTimeoutInSeconds: number;
     public readonly automaticSilentRenew: boolean;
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oidc-client-ts",
-  "version": "2.0.3",
```
pamapa, line 3: Please do not touch package-lock.json and package.json in your MR.

.gitignore
```diff
@@ -21,3 +21,4 @@ node_modules/
 .DS_Store
 .vscode/
 temp/
+.history/
```
pamapa, line 24: i do not see why this would be needed

pavliy: I'm using localhistory plugin for vscode (https://marketplace.visualstudio.com/items?itemName=xyz.local-history)
So that's just to ignore files generated by it.

package.json
```diff
@@ -37,6 +37,7 @@
     "clean": "git clean -fdX dist lib *.tsbuildinfo",
     "prepack": "npm run build",
     "test": "tsc && jest",
+    "test:watch": "npm run test -- --watch",
```
pamapa, line 40: if you want this please make another MR with only that

pamapa:
Please rebase all commits into one commit (as from different users...). Otherwise LGTM.

@kherock Can you have a look too?

pavliy:
> Please rebase all commits into one commit (as from different users...). Otherwise LGTM.
> 
Will do. Thnx

pavliy:
@pamapa @kherock please let me know if anything else is needed in this MR.
thnx

pamapa:
i have added one comment to fix, if this is done i will merge this unless @kherock as more to add...

src/UserManagerSettings.ts
```diff
@@ -36,6 +36,10 @@ export interface UserManagerSettings extends OidcClientSettings {
     /** The methods window.location method used to redirect (default: "assign") */
     redirectMethod?: "replace" | "assign";
 
+    /** The target to pass while calling postMessage inside iframe for callback */
```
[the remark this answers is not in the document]
pavliy: Makes sense. thnx.

pamapa (approved):

pamapa:

PR #535, opened 2022-05-18
origin based URL fix #534

### LINKED ISSUE #534: signinRedirectCallback should accept a relative path
Hi,

first THANK YOU for this port. I would like to report a small inconvenience.

```ts
    signinRedirectCallback(url?: string): Promise<User>;
```
only accepts full URL, including scheme and host.

Vue router guards (i.e. middleware) parameters only gives the relative path (cf. https://router.vuejs.org/api/#fullpath). Scheme and host are never provided by Vue router in order to enforce portability across environments.

I had to change from :
```ts
function endLogin(to, from, next, userManager) {
  return userManager
    .signinRedirectCallback(to.fullPath)
    .then(user => ...)
}
```
To something like
```ts
function endLogin(to, from, next, userManager) {
  const origin = new URL(process.env.VUE_APP_CALLBACK_ROOT_URL).origin;
  const fullUrl = origin + to.fullPath;
  return userManager
    .signinRedirectCallback(fullUrl)
    .then(user => ...)
}
```
where `VUE_APP_CALLBACK_ROOT_URL` is an environment variable also used for `UserManager({ redirect_uri })` setting. I found the former more readable.

Thanks again

ydarma:
fixes #534

### Checklist

- [ ] This PR makes changes to the public API : no

### DIFF
--- a/src/utils/UrlUtils.ts
+++ b/src/utils/UrlUtils.ts
@@ -6,7 +6,7 @@
  */
 export class UrlUtils {
     public static readParams(url: string, responseMode: "query" | "fragment" = "query"): URLSearchParams {
-        if (!url) throw new Error("Invalid URL");
+        if (!url) throw new TypeError("Invalid URL");
         const parsedUrl = new URL(url, window.location.origin);
         const params = parsedUrl[responseMode === "fragment" ? "hash" : "search"];
         return new URLSearchParams(params.slice(1));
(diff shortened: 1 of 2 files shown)


src/utils/UrlUtils.ts
```diff
@@ -6,7 +6,7 @@
  */
 export class UrlUtils {
     public static readParams(url: string, responseMode: "query" | "fragment" = "query"): URLSearchParams {
-        if (!url) throw "Invalid URL";
+        if (!url) throw new Error("Invalid URL");
```
pamapa, line 9: thanks, i updated my previous comment, maybe i should have made a second one. What i edited:
Its probably better to use `TypeError` instead of `Error`, as this is what `new URL` would do. Sorry for the trouble. Should be easy to fix :-)

pamapa (approved):

pamapa:
thanks for contributing!

PR #538, opened 2022-05-20
allow override of prompt on signin

### LINKED ISSUE #536: support overriding "prompt" on signinRedirect
I have two calls to signinRedirect. One does not pass prompt, the other passes prompt=login so that I can force the external provider to show the login prompt. Looks like you can do this, but the SignInRedirectArgs don't currently support it. The old oidc-client allowed you to override pretty much everthing so I guess there are probably more things people will want to override going forward.

justin-mellor:
Closes/fixes #536

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -112,7 +112,7 @@ export class ErrorTimeout extends Error {
 }
 
 // @public (undocumented)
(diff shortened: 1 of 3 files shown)


pamapa (approved):

pamapa:
thanks for contributing

PR #570, opened 2022-06-10
preserve session_state during refresh token

### LINKED ISSUE #569: session_state is lost on silent refresh token renewal
I am trying to use the silent refresh feature with session state management, but run into the following issue:

I have very short access token lifetime (5min). When oidc-client-ts (successfully, I might add) refreshes the access token using the refresh token, it seems to clear the `session_state` field of User. This leads to session management not working when I reload the page, as now `session_state` is null.

Steps to reproduce:
- Set up a client with `monitorSession: true`, short access token lifetimes and refresh tokens
- Log in without any previous state
- Observe that the session mgmt iframe is loaded (via the debug log for example)
- Wait for silent refresh token renewal
- Reload the page
- Observe that no login attempt is made (since there is still a valid access token in sessionStorage)
- Observe that the session mgmt iframe is *not* loaded (since session_state is now null)
- This can also be verified by checking the content of sessionStorage

pamapa:
Closes/fixes #569

### Checklist

### DIFF
--- a/src/ResponseValidator.ts
+++ b/src/ResponseValidator.ts
@@ -75,6 +75,8 @@ export class ResponseValidator {
         const logger = this._logger.create("validateRefreshResponse");
 
         response.userState = state.data;
+        // if there's no session_state on the response, copy over session_state from original request
+        response.session_state ??= state.session_state;
         // if there's no scope on the response, then assume all scopes granted (per-spec) and copy over scopes from original request
         response.scope ??= state.scope;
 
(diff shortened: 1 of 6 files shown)


src/ResponseValidator.ts
```diff
@@ -75,6 +75,8 @@ export class ResponseValidator {
         const logger = this._logger.create("validateRefreshResponse");
 
         response.userState = state.data;
+        // if there's no session_state on the response, copy over scopes from original request
```
kherock, line 78: The word "scopes" shouldn't be here

kherock (approved):
Thanks @pamapa!

PR #613, opened 2022-07-08
Add support for OIDC session management error status

### LINKED ISSUE #611: OIDC session management error status not supported
Currently oidc-client-ts does not support the status `error` that is specified in the OIDC session management specification (https://openid.net/specs/openid-connect-session-1_0.html section 3). The error status is supposed to be the result of any error that occurs during the processing of the `session_state` in the OP iFrame.

This feature was also not supported in the "original" JavaScript version of this library. I had created a pull request to add the support for the error status but the maintainers of the JavaScript library did not merge the request before deciding to discontinue the development. Therefore I will provide the PR again to add the support of the error status to this library.

divid3byzero:
Closes/fixes #611 

### Checklist

### DIFF
--- a/src/CheckSessionIFrame.ts
+++ b/src/CheckSessionIFrame.ts
@@ -19,6 +19,8 @@ export class CheckSessionIFrame {
         url: string,
         private _intervalInSeconds: number,
         private _stopOnError: boolean,
+        private propagateUserSessionError: boolean,
+        private userSessionErrorCallback: () => void,
     ) {
         const parsedUrl = new URL(url);
         this._frame_origin = parsedUrl.origin;
@@ -54,6 +56,10 @@ export class CheckSessionIFrame {
                 this._logger.error("error message from check session op iframe");
                 if (this._stopOnError) {
                     this.stop();
+                }
+
+                if (this.propagateUserSessionError) {
+                    this.userSessionErrorCallback();
                 }
             }
             else if (e.data === "changed") {
--- a/src/SessionMonitor.ts
+++ b/src/SessionMonitor.ts
@@ -37,8 +37,7 @@ export class SessionMonitor {
         // doesn't trigger load event.
         if (user) {
             void this._start(user);
-        }
-        else if (this._userManager.settings.monitorAnonymousSession) {
+        } else if (this._userManager.settings.monitorAnonymousSession) {
             const session = await this._userManager.querySessionStatus();
             if (session) {
                 const tmpUser = {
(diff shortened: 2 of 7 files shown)


pamapa:
`_raiseUserSessionError` is never called. What is the difference to `_userSessionChanged`. Why can't you use that one?

divid3byzero:
You are right, sorry, I didn't commit everything. I messed up my local development and forgot a few things in the commit after fixing it. I've committed the missing code.

Actually there are use cases when clients want to react to the error status. And as that is something different than the changed status I (and everyone else that wants to react differently to the two statuses) can't use `userSessionChanged`.

src/CheckSessionIFrame.ts
```diff
@@ -52,9 +54,13 @@ export class CheckSessionIFrame {
         ) {
             if (e.data === "error") {
                 this._logger.error("error message from check session op iframe");
-                if (this._stopOnError) {
+                if (this._stopOnError && !this.propagateUserSessionError) {
```
pamapa, line 57: if `_stopOnError` is true i really would expect that `this.stop()` should be called. You need to rearrange your code changes

src/CheckSessionIFrame.ts
```diff
@@ -55,6 +57,10 @@ export class CheckSessionIFrame {
                 if (this._stopOnError) {
                     this.stop();
                 }
+
+                if (this.propagateUserSessionError) {
```
pamapa, line 61: I guess we could combine `propagateUserSessionError` and `userSessionErrorCallback`: Make `userSessionErrorCallback` optional, if defined call it if not not. Then `propagateUserSessionError`  is not needed inside this class...

divid3byzero: Actually the way I implemented it is the same pattern as the `_callback` is called after the `changed` status is triggered. By using the `propagateUserSessionError ` flag it is made optional. If you'd rather like that the existence of the `userSessionErrorCallback` is checked to trigger that callback I would have to allow for a function to be defined in the settings object which, of course, would be possible but would break with the way that a client can attach to e.g. the `changed` status.

src/SessionMonitor.ts
```diff
@@ -90,21 +88,27 @@ export class SessionMonitor {
...
             }
-        }
-        catch (err) {
+        } catch (err) {
             // catch to suppress errors since we're in non-promise callback
             logger.error("Error from getCheckSessionIframe:", err instanceof Error ? err.message : err);
         }
     };
 
+    private doUserSessionErrorPropagation(): boolean {
```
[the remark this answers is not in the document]
divid3byzero: This is a matter of preference, but yes, I will introduce a variable.

phlegx:
Nice work!

PR #612, opened 2022-07-08
Add support for id_token_hint parameter in prompt=none request

divid3byzero:
Closes/fixes #issue #580 

### Checklist

- [x] This PR makes changes to the public API 
- [x] I have included links for closing relevant issue numbers

PR #622, opened 2022-07-15
add id_token_hint if present

### LINKED ISSUE #580: Configuration of parameter includeIdTokenInSilentRenew has no effect
As per OIDC specification (https://openid.net/specs/openid-connect-session-1_0.html) an RP should send the current ID token of a user when attempting an silent renew via `prompt=none` call. To send the ID token using the id_token_hint query parameter as described in the specification the documentation states that the `UserManagerSetting` parameter `includeIdTokenInSilentRenew ` should be set to `true` (it is `false` by default). It appears, though, that this configuration parameter has no effect as the ID token is not appended to the `prompt=none` call resulting in an error message from the OP. 

On checking the code I found that the OIDC client configuration ist passed to the class `UserManagerSettingsStore` (https://github.com/authts/oidc-client-ts/blob/main/src/UserManagerSettings.ts) where the configuration parameters are assigned using the values set in the OIDC client. It seems that in the case of the parameter `includeIdTokenInSilentRenew` the value is hard coded to `false` and any other value that is configured using the settings object is simply ignored and the default value of `false` can thus never be overridden. 

I need to be able to set this parameter to `true` as the OP I am using will refuse to process any `promp=none` calls without the `id_token_hint` parameter. Just to be clear: this behaviour of the OP is correct, it behaves exactly as stated in the specification. So this is something that needs to be fixed on the client (library) side. I can provide a pull request if whished.

I would also be glad if someone could point me in the right direction in case I am simply using the library in the wrong way or if I've overlooked anything.

pamapa:
Closes #580

- Take over from MR #612 as user has problems with formatting/git
- For now i have bound that feature to `includeIdTokenInSilentRenew` as noted in the related issue #580, if that leads to a problem for others we can still add an additional setting option.

### Checklist

### DIFF
--- a/src/UserManager.ts
+++ b/src/UserManager.ts
@@ -329,11 +329,13 @@ export class UserManager {
             logger.throw(new Error("No silent_redirect_uri configured"));
         }
 
+        const user = await this._loadUser();
         const handle = await this._iframeNavigator.prepare({ silentRequestTimeoutInSeconds });
         const navResponse = await this._signinStart({
             request_type: "si:s", // this acts like a signin silent
             redirect_uri: url,
             prompt: "none",
+            id_token_hint: this.settings.includeIdTokenInSilentRenew ? user?.id_token : undefined,
             response_type: this.settings.query_status_response_type,
             scope: "openid",
             skipUserInfo: true,


pamapa:
@divid3byzero Please test and review

src/UserManager.ts
```diff
@@ -328,12 +328,14 @@ export class UserManager {
         if (!url) {
             logger.throw(new Error("No silent_redirect_uri configured"));
         }
+        const user: User | null = await this.getUser();
```
longsleep, line 331: To avoid triggering an "user loaded" this should use `this._loadUser()` here as the only purpose of retrieving the current user is to provide the `id_token_hint`.

longsleep (changes requested):

src/UserManager.ts
pamapa: good point, fixed!

longsleep (approved):
LGTM now :+1:

Issue #607, opened 2022-07-05
Pass cookies when renewing access token


thomasmikava:
Is it possible to include cookies in the request when any request (such as access token renewal) is sent? It should be possible by passing `{ credentials: 'include' }` to `fetch` function, but currently it's not possible to specify extra settings to `fetch`.

pamapa:
related to #608

PR #659, opened 2022-08-30
new option refreshWithCredentials sends cookies with token refresh

oberhamsi:
fixes https://github.com/authts/oidc-client-ts/issues/607

PR could be improved by allowing more `credentials` settings besides include.

the PR is very helpful in the case that the openid server sets cookies which can be updated by refresh so that silent signin works longer.

### Checklist

- [x ] This PR makes changes to the public API 
- [ x] I have included links for closing relevant issue numbers

### DIFF
--- a/src/JsonService.ts
+++ b/src/JsonService.ts
@@ -23,6 +23,7 @@ export interface PostFormOpts {
     body: URLSearchParams;
     basicAuth?: string;
     timeoutInSeconds?: number;
+    withCredentials?: boolean;
 }
 
 /**
@@ -123,6 +124,7 @@ export class JsonService {
         body,
         basicAuth,
         timeoutInSeconds,
+        withCredentials,
     }: PostFormOpts): Promise<Record<string, unknown>> {
         const logger = this._logger.create("postForm");
         const headers: HeadersInit = {
@@ -132,11 +134,15 @@ export class JsonService {
         if (basicAuth !== undefined) {
             headers["Authorization"] = "Basic " + basicAuth;
         }
+        let initCredentials = {};
+        if (withCredentials) {
+            initCredentials = { "credentials": "include" };
+        }
 
         let response: Response;
         try {
             logger.debug("url:", url);
-            response = await this.fetchWithTimeout(url, { method: "POST", headers, body, timeoutInSeconds });
+            response = await this.fetchWithTimeout(url, { method: "POST", headers, body, timeoutInSeconds, ...initCredentials });
         }
         catch (err) {
             logger.error("Network error");
--- a/src/JsonService.ts
+++ b/src/JsonService.ts
@@ -23,7 +23,7 @@ export interface PostFormOpts {
     body: URLSearchParams;
     basicAuth?: string;
     timeoutInSeconds?: number;
-    corsCredentials?: "same-origin" | "include" | "omit";
+    initCredentials?: "same-origin" | "include" | "omit";
 }
 
 /**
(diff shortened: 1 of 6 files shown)


pamapa:
- Test fails
- would be nice if as you noted more options would be supported
  - "same-origin" (default)
  - "include"
  - "omit"
- the settings name could be more explicit
  - refreshTokenCredentials?: "same-origin" | "include" | "omit"

oberhamsi:
thanks for the quick review, i fixed the unit test and changed it to `refreshTokenCredentials` and allow all possible options

src/JsonService.ts
```diff
@@ -23,6 +23,7 @@ export interface PostFormOpts {
     body: URLSearchParams;
     basicAuth?: string;
     timeoutInSeconds?: number;
+    corsCredentials?: "same-origin" | "include" | "omit";
```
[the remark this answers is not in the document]
oberhamsi: i simplified the initCredentials creation. i can rename `corsCredentials` too but i don't see the benefit

src/OidcClientSettings.ts
```diff
@@ -102,6 +102,11 @@ export interface OidcClientSettings {
     extraQueryParams?: Record<string, string | number | boolean>;
 
     extraTokenParams?: Record<string, unknown>;
+
+    /**
+     * CORS settings for credentials sent with the refresh request.
```
pamapa, line 107: Improve the documentation:
`Credentials used by fetch with the refresh request. (default: "same-origin")`

I guess its better to explicitly set `refreshTokenCredentials` to "same-origin", to avoid possible but unlikely different browser defaults...

src/OidcClientSettings.ts
```diff
@@ -168,6 +175,7 @@ export class OidcClientSettingsStore {
         mergeClaims = false,
         // other behavior
         stateStore,
+        refreshTokenCredentials,
```
pamapa, line 178: refreshTokenCredentials = "same-origin"

See comment above, why...

oberhamsi: yes, will do. though about this too :)

src/OidcClientSettings.ts
```diff
@@ -150,6 +155,8 @@ export class OidcClientSettingsStore {
     public readonly extraQueryParams: Record<string, string | number | boolean>;
     public readonly extraTokenParams: Record<string, unknown>;
 
+    public readonly refreshTokenCredentials: "same-origin" | "include" | "omit" | undefined;
```
pamapa, line 158: no undefined needed

src/TokenClient.ts
```diff
@@ -31,6 +31,8 @@ export interface ExchangeRefreshTokenArgs {
     scope?: string;
 
     timeoutInSeconds?: number;
+
```
pamapa, line 34: avoid newline here

src/OidcClientSettings.ts
oberhamsi: defaulting to same-origin in OidcClientSettingsStore

src/JsonService.ts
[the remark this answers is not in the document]
pamapa: it has only indirectly to do with cors, e.g. "omit" will not send any cookie at all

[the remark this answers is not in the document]
oberhamsi: ah yes, i changed it to `initCredentials`

pamapa:
Looks good, ready to merge? I will squash all commit into one unless you rebase first :-) ...

pamapa (approved):

oberhamsi:
Yes, looks good to me. Merge at your pleasure

pamapa:
Nice work, thanks for contributing!

PR #690, opened 2022-09-16
fix id_token is lost on silent refresh token renewal

### LINKED ISSUE #688: User profile information is lost when refresh token is used.
Hi!

I've run into an issue where the profile object on my User is empty after the refresh token is used to get a new access token. It looks like useRefreshToken in the client generates an empty SigninResponse and merges in the new access and refresh tokens to that object. That causes an empty profile to get returned when the new User object is created. Should we maybe include the profile in the RefreshState object so it can be merged into the new User? Or am I just doing something incorrectly. Thanks for the help!

pamapa:
Closes/fixes #688

### Checklist

### DIFF
--- a/src/ResponseValidator.ts
+++ b/src/ResponseValidator.ts
@@ -82,12 +82,20 @@ export class ResponseValidator {
 
         // OpenID Connect Core 1.0 says that id_token is optional in refresh response:
         // https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse
-        const hasIdToken = response.isOpenId && !!response.id_token;
-        if (hasIdToken) {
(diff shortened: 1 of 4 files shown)


src/ResponseValidator.ts
```diff
@@ -87,6 +87,10 @@ export class ResponseValidator {
             this._validateIdTokenAttributes(response, state.id_token);
             logger.debug("ID Token validated");
         }
+        else if (state.id_token) {
```
pamapa, line 90: Maybe ts better to pass user.profile into state and reuse that, as it might contains more info than id_token. Profile is a combination of the `id_token` and the user info endpoint...

longsleep (approved):
LGTM :+1:

PR #736, opened 2022-10-25
fix(oidcclient):for issue 734

### LINKED ISSUE #734: using hash router, getting the parameters from the url is incorrect
Situation:
When the router is set to hashmode the parameters are not correctly parsed from the url.
and you get the error that the state is not found.

Analysing:
- UrlUtils.readParams expects an unhashed routed url

easiest way yo fix: 
const unHashUrl = url.replace('#/', '');
const response = new SigninResponse(UrlUtils.readParams(unHashUrl, this.settings.response_mode));

and 

const unHashUrl = url.replace('#/', '');
const response = new SignoutResponse(UrlUtils.readParams(unHashUrl, this.settings.response_mode));

frontendplace:
Closes/fixes #734

### Checklist

### DIFF
--- a/src/utils/UrlUtils.test.ts
+++ b/src/utils/UrlUtils.test.ts
@@ -34,6 +34,16 @@ describe("UrlUtils", () => {
             expect(resultObj).toHaveProperty("foo", "test");
         });
 
+        it("should return query params when hash routes are used", () => {
+            // act
+            const result = UrlUtils.readParams("http://app/#/path?code=foo&status=bar");
+            const resultObj = Object.fromEntries(result);
+
+            // assert
+            expect(resultObj).toHaveProperty("code", "foo");
+            expect(resultObj).toHaveProperty("status", "bar");
+        });
+
         it("should return fragment params for response_mode=fragment when path is relative", () => {
             // act
             const result = UrlUtils.readParams("/app/?foo=test#bar=test_fragment", "fragment");
--- a/src/utils/UrlUtils.ts
+++ b/src/utils/UrlUtils.ts
@@ -9,7 +9,7 @@ export class UrlUtils {
         if (!url) throw new TypeError("Invalid URL");
 
         // if we're doing query, then strip off hash fragment before we parse
-        if (responseMode == "query") {
+        if (responseMode === "query") {
             const hashIdx = url.indexOf("#");
             if (hashIdx >= 0) {
                 url = url.substr(hashIdx + 1);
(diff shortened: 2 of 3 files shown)


src/utils/UrlUtils.ts
```diff
@@ -7,6 +7,15 @@
 export class UrlUtils {
     public static readParams(url: string, responseMode: "query" | "fragment" = "query"): URLSearchParams {
         if (!url) throw new TypeError("Invalid URL");
+
+        // if we're doing query, then strip off hash fragment before we parse
+        if (responseMode === "query") {
+            const hashIdx = url.indexOf("#");
+            if (hashIdx >= 0) {
+                url = url.substr(hashIdx + 1);
```
pamapa, line 15: please replace `substr` with not deprecated `substring` version

src/utils/UrlUtils.ts
```diff
@@ -7,6 +7,15 @@
...
+            }
+        }
+
         const parsedUrl = new URL(url, window.location.origin);
```
pamapa, line 19: What about the baseUrl `window.location.origin`? Do we need to strip here as well?

pamapa:
Thanks for taking care!

src/utils/UrlUtils.ts
frontendplace: no the origin doesn't have a hash

src/utils/UrlUtils.ts
frontendplace: done

pamapa:
@longsleep Can you please have a look too?

src/utils/UrlUtils.test.ts
```diff
@@ -34,6 +34,16 @@ describe("UrlUtils", () => {
             expect(resultObj).toHaveProperty("foo", "test");
         });
 
+        it("should return query params when hash routes are used", () => {
+            // act
+            const result = UrlUtils.readParams("http://app/#/path?code=foo&status=bar");
```
longsleep, line 39: This is not the correct way to define urls with query parameter and imo it is not our job to deal with this..

Everything after the `#` is part of the hash and that includes the `?` etc. Means in technical terms this URL has no query.

Only the application specific hash based router knows about this and thus it must do URL parsing before our fallback is run.

To fix the test, please put the ? and it's values before the # so we ensure that this works (removing the change from below).

longsleep (changes requested):
This can cause undesired results as imo the assumptions which are base for this pr are wrong. Should probably discussed further in the associated issue

frontendplace:
Ill remove the PR. 
I can see that I can pass a modified url in angular when calling the signinRedirectCallback function

src/utils/UrlUtils.test.ts
frontendplace: De PR can be deleted I see I can fix this in our angular Code by passing a modified version of the default window.location.href

Issue #665, opened 2022-09-01
UserManager support Password Credentials Grant


wenerme:
Password Credentials Grant https://www.rfc-editor.org/rfc/rfc6749#section-4.3

Just a post

```http
POST https://keycloak/auth/realms/demo/protocol/openid-connect/token
Content-Type: application/x-www-form-urlencoded

grant_type=password&username=admin&password=admin&client_id=demo-web
```

wenerme:
A workaround 

```ts
class ClientAuthUserManager extends UserManager {
  async signinCredential({ username, password }: { username: string; password: string }) {
    let logger = this._logger.create('signinCredential');
    let url = await this.metadataService.getTokenEndpoint();
    if (!url) {
      throw new Error(``);
    }
    let resp;
    try {
      let r = await fetch(url, {
        method: 'POST',
        headers: {
          'Content-Type': 'application/x-www-form-urlencoded',
        },
        body: new URLSearchParams({
          username,
          password,
          grant_type: 'password',
          client_id: this.settings.client_id,
        }).toString(),
        mode: 'cors',
      });
      resp = await r.json();
    } catch (e) {
      logger.error(`Login failed`, e);
      throw new Error('Login failed');
    }

    let user = new User(resp);
    await this.storeUser(user);
    logger.debug('user stored');
    this._events.load(user);
    return user;
  }
}
```

pamapa:
Thanks for sharing, i would accept a MR, which adds this feature. Most of the part can be achieved with `JsonService.post` i guess...

jsedanoj:
@wenerme :

There is another similar issue from me (https://github.com/authts/oidc-client-ts/issues/746) and a PR (https://github.com/authts/oidc-client-ts/pull/747). Feel free to comment if it does not fit your needs and I'll do my best.

jsedanoj:
I think we should close this issue.

PR #747, opened 2022-11-01
Add Resource Owner Password Credentials flow

### LINKED ISSUE #746: PoC support for Resource Owner Password Credential Flow
Hello,

I need support for Resource Owner Password Credential Flow (https://www.rfc-editor.org/rfc/rfc6749#section-4.3). `oidc-client-ts` does not support it... but the changes needed to support it are very small.

I have made it in my OIDC playground, but I have needed to duplicate some classes in my project because they are @internal and cannot be used/extended in client projects. It would be much easier if this support was directly provided by oidc-client-ts. Watch my PoC at https://gitlab.com/javier-sedano/oidc-vue/-/tree/master/src/oidc-client-ts and deployed to "production" at https://jsedano.duckdns.org/oidc-vue/ .

My questions:

- Are you guys open to receiving such a contribution?
- How? Fork + PR?

BTW, I know that Resource Owner Password Credential Flow is only secure if the Client Application is fully trusted not to keep the credentials... but in scenarios where the IdP and the Client Application are owned by the same entity (which is my case) it is acceptable (for example, read https://www.rfc-editor.org/rfc/rfc6749#section-1.3.3 or  https://auth0.com/docs/get-started/authentication-and-authorization-flow/resource-owner-password-flow). So I would add the appropriate warning to the readme.

jsedanoj:
Closes #746 and #665

I have done my best to mimic the patterns and name conventions in the existing code base, but comments are welcome.

### Checklist


pamapa:
Excellent work, small pipeline issue:
```
oidc-client-ts@2.1.0 lint


/home/runner/work/oidc-client-ts/oidc-client-ts/src/ResponseValidator.test.ts
Warning:   4:20  warning  'Timer' is defined but never used  @typescript-eslint/no-unused-vars
ESLint found too many warnings (maximum: 0).
```

jsedanoj:
> Excellent work, small pipeline issue:
> 
> ```
> oidc-client-ts@2.1.0 lint
> > eslint --max-warnings=0 --cache .
> 
> 
> /home/runner/work/oidc-client-ts/oidc-client-ts/src/ResponseValidator.test.ts
> Warning:   4:20  warning  'Timer' is defined but never used  @typescript-eslint/no-unused-vars
> ESLint found too many warnings (maximum: 0).
> ```

Thank you. Fixed.

README.md
```diff
@@ -24,6 +24,7 @@ Implements the following OAuth 2.0 protocols and supports
 
 - [Authorization Code Grant](https://oauth.net/2/grant-types/authorization-code/)
   with [PKCE](https://oauth.net/2/pkce/)
+- [Resource Owner Password Credentials Grant](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.3); This flow must be used only when the Client Application is fully trusted by the Authorization Server / Identity Provider; [See more info](https://auth0.com/docs/get-started/authentication-and-authorization-flow/resource-owner-password-flow)
```
longsleep, line 27: @jsedanoj i am curious under what circumstances an application which uses oidc-client-ts can be considered fully trusted.

Would you please share some details and maybe add a section about how to ensure this so that people do not use this flow unless in very particular use cases.

I am a bit doubtful if we really should add something which for example would enable people to simply store the username/password in LocalStorage - or simply speaking - allow flows which are outside of the OpenID connect use case in general.

longsleep (commented):
I don't have a strong opinion on if or not mege of this - but I think we should carefully evaluate if it's a good idea to open up this library for potentially outdated / legacy use cases.

src/OidcClient.ts
```diff
@@ -165,6 +165,18 @@ export class OidcClient {
         return response;
     }
 
+    public async processResourceOwnerPasswordCredentials(
```
pamapa, line 168: would be nice to have the params packed into one object:
```
/**
 * @public
 */
export type ProcessResourceOwnerPasswordCredentialsArgs = {
    username: string;
    password: string;
    skipUserInfo?: boolean;
};
...
    public async processResourceOwnerPasswordCredentials({
        username: string,
        password: string,
        skipUserInfo: boolean
    }: ProcessResourceOwnerPasswordCredentialsArgs): Promise<SigninResponse> {
````

File: src/UserManager.ts [hunk 3]
```diff
@@ -404,6 +439,13 @@ export class UserManager {
         const signinResponse = await this._client.processSigninResponse(url);
         logger.debug("got signin response");
 
+        const user = await this._buildUser(signinResponse, verifySub);
+
```
pamapa, line 443: no newline needed here

src/UserManager.ts
```diff
@@ -171,6 +181,31 @@ export class UserManager {
...
+        password,
+        skipUserInfo = false,
+    }: SigninCredentialsArgs) {
+        const logger = this._logger.create("signinResourceOwnerCredential");
+
+        const signinResponse = await this._client.processResourceOwnerPasswordCredentials(username, password, skipUserInfo);
+        logger.debug("got signin response");
+
+        const user = await this._buildUser(signinResponse);
+
```
pamapa, line 200: no newline needed here

src/UserManager.ts
```diff
@@ -45,6 +46,15 @@ export type SigninPopupArgs = PopupWindowParams & ExtraSigninRequestArgs;
  */
 export type SigninSilentArgs = IFrameWindowParams & ExtraSigninRequestArgs;
 
+/**
+ * @public
+ */
+export type SigninCredentialsArgs = {
```
pamapa, line 52: now we can reuse the type from OidcClient.ts and do a small rename too:
```export type SigninResourceOwnerCredentialsArgs = ProcessResourceOwnerPasswordCredentialsArgs;```

File: README.md (hunk 1 above)
Comment (pamapa) [reply README.md:27]: I agree, we should add more precise text here!

From the auth0 page:
```
Because credentials are sent to the backend and can be stored for future use before being exchanged for an Access Token, it is imperative that the application is absolutely trusted with this information. 
```

This flow can only be used when the client is **not** exposed to the user. When the client is running within the users browser, this flow is a no go, as it can not protect the credentials...

Comment (jsedanoj) [reply README.md:27]: The RFC (https://www.rfc-editor.org/rfc/rfc6749#section-4.3) says "The client MUST discard the credentials once an access token has been obtained". So a fully trusted client is a client you trust to follow the RFC (i.e.: do not keep the user's credentials... which is your 3rd paragraph); mainly, when the IdP and the client application are developed by the same company. Still, even in this case, the attack surface is increased. I can make this statement more explicit in the README if you prefer.

My use case is the following: the company is deploying an IdP. We want all of our applications to use the IdP to have one single source of truth. For some applications (backoffice applications), the redirection to the IdP is more than acceptable. But for some frontoffice applications, the UX may be weird if we do that (end users are not used to that). Hence, the Resource Owner Password Credentials flow is the solution. Notice that the client application is developed by the same company, so we can audit the application and ensure that it is not storing the credentials.

However, you are right in a sense: it is not part of the OAuth 2.1 draft (https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-07#section-10), which is what oidc-client-ts explicitly says to focus on.

If the PR is going to be rejected (and I can understand the reasons, that's why I asked before doing it https://github.com/authts/oidc-client-ts/issues/746), can we consider another solution: make the needed changes so someone willing to use this flow can extend the classes and do it himself in his project. This would imply:

- Expose some inner classes (mainly JsonService), so this other proposal (https://github.com/authts/oidc-client-ts/issues/665) can reuse them.

If this is the way to go, I can also try to do this PR.

Comment (jsedanoj) [reply README.md:27]: > I agree, we should add more precise text here!
> 
> From the auth0 page:
> 
> ```
> Because credentials are sent to the backend and can be stored for future use before
> being exchanged for an Access Token, it is imperative that the application is absolutely
> trusted with this information. 
> ```
> 
> This flow can only be used when the client is **not** exposed to the user. When the client is running within the users browser, this flow is a no go, as it can not protect the credentials...

I disagree with the last paragraph. If the application is exposed to the user (running in the browser), yes, the credentials are exposed to the user... but they are HIS credentials. This is not a big deal.

For me, there are three concerns with this flow:

- It is not part of the OAuth 2.1 spec, which oidc-client-ts says to adhere to. It is part of OAuth 2.0, though.

- It is only usable if the Client Application is trusted not to keep the credentials (so developed by the same organization, or very similar constraints).

- Still, it increases the attack surface. When using the Authorization Code flow, if the Client Application is compromised (for example XSS), the access_token is compromised, so the resources protected by this access_token are compromised. When using Resource Owner Password Credentials flow, if the Client Application is compromised, the credentials of the user are compromised, so other applications accessible by the same user are compromised as well.

Comment (pamapa) [reply README.md:27]: I am fine to have this flow even so it is not part of the "OAuth 2.1 draft". Because it fits in this library quiet nicely and its better to have it inside than having multiple implementations outside...

But please update the README.md description, such that is more clear for the users when to use this flow and when not.

Comment (longsleep) [reply README.md:27]: @jsedanoj 

>  the UX may be weird if we do that (end users are not used to that). Hence, the Resource Owner Password Credentials flow is the solution.

This is a common problem and its also what feared, since you can still do a secure code flow even in such cases by using `prompt=none` and have some way to figure out if a client is trusted or not (for example via client registration, to avoid interactive prompt).

I can only strongly advise to avoid any flow which already is in a web browser to use flows which directly use the credentials as it is almost never needed.

Comment (jsedanoj) [reply README.md:27]: > @jsedanoj
> 
> > the UX may be weird if we do that (end users are not used to that). Hence, the Resource Owner Password Credentials flow is the solution.
> 
> This is a common problem and its also what feared, since you can still do a secure code flow even in such cases by using `prompt=none` and have some way to figure out if a client is trusted or not (for example via client registration, to avoid interactive prompt).
> 

Sorry, I do not understand this paragraph. Let me explain what I understand and correct me if I misunderstood.

With prompt=none, we are telling the IdP that he must not show any prompt to the user. So the most common cases are:
- If the user is previously authenticated in the IdP (session cookie or whatever other way t is using), the user will be redirected to the IdP, redirected back to the Client and everything is ok. For the user, it is mostly transparent, almost like if nothing has happened, except because there is a reload of the page (which, in a modern SPA, may be UX problem by itself, but well, that's life).
- If the user is not previously authenticated in the IdP (1)... the authentication fails.

So I do not see how the prompt=none helps the UX... unless... I think we are talking about different UXs here. The UX I am talking about is: when a user is login into something.com, he expects to put his credentials in something.com, not in idp.something.com with a reload . In a modern SPA maybe even the behaviour of the application changes without a page reload. So the login form is directly inside the application, not in a redirect. See this example:

image

(1) There may be other reasons, such as "the user is authenticated, but he has not given grant to the Client Application).

> I can only strongly advise to avoid any flow which already is in a web browser to use flows which directly use the credentials as it is almost never needed.

I also advise against it. Having the Client application manage the credentials means that even if you make the IdP a very secure system... if you have a security problem in the Client application, then the full user is exposed.

But sometimes my advise conflicts the $ (a 1% enhancement in the acquisition funnel is a lot of $$$$$), and I prefer a solution 50% battle-tested than a from-scratch implementation.

Comment (longsleep) [reply README.md:27]: > If the user is previously authenticated in the IdP (session cookie or whatever other way t is using), the user will be redirected to the IdP, redirected back to the Client and everything is ok. For the user, it is mostly transparent, almost like if nothing has happened, except because there is a reload of the page (which, in a modern SPA, may be UX problem by itself, but well, that's life).

With `prompt=none` one usually does not redirect the user visible page. Instead this is done in an hidden iframe (this is called silent-refresh in oidc-client-ts). Only if silent refresh fails one gets redirected away. And based on the IdP even that can be avoided (all possible if you control both IdP and app), by rendering the interactive authorization in an this time visible iframe or in an popup (some IdPs support the display query parameter to further customize the looks so that this works nicely and invisible to the user).

> I am talking about is: when a user is login into something.com, he expects to put his credentials in something.com, not in idp.something.com with a reload . In a modern SPA maybe even the behaviour of the application changes without a page reload.

Well run the IdP on the same URL should not be a problem. What you say is exactly one of the anti-pattern why modern auth flows were invented, namely in this case exactly to avoid the user credentials like passwords etc flowing around to multiple places and even getting relayed. Instead there they should flow to exactly the IdP which then offers some sort of grant (via token).

Comment (jsedanoj) [reply README.md:27]: > > If the user is previously authenticated in the IdP (session cookie or whatever other way t is using), the user will be redirected to the IdP, redirected back to the Client and everything is ok. For the user, it is mostly transparent, almost like if nothing has happened, except because there is a reload of the page (which, in a modern SPA, may be UX problem by itself, but well, that's life).
> 
> With `prompt=none` one usually does not redirect the user visible page. Instead this is done in an hidden iframe (this is called silent-refresh in oidc-client-ts). Only if silent refresh fails one gets redirected away. And based on the IdP even that can be avoided (all possible if you control both IdP and app), by rendering the interactive authorization in an this time visible iframe or in an popup (some IdPs support the display query parameter to further customize the looks so that this works nicely and invisible to the user).

Wouldn't this be equally risky? If the Client is compromised (through for example an XSS attack), the iframe is compromised as well... No?

Do we have an example of this, anyway? I control the IdP (based on KeyCloak).

> 
> > I am talking about is: when a user is login into something.com, he expects to put his credentials in something.com, not in idp.something.com with a reload . In a modern SPA maybe even the behaviour of the application changes without a page reload.
> 
> Well run the IdP on the same URL should not be a problem. What you say is exactly one of the anti-pattern why modern auth flows were invented, namely in this case exactly to avoid the user credentials like passwords etc flowing around to multiple places and even getting relayed. Instead there they should flow to exactly the IdP which then offers some sort of grant (via token).

The UX problem is not the fact that it is in a different URL (most users will not read the URL anyway): it is the fact that it causes a reload. But if we have an example using the signinSilent() and filling iframe with the IdP login form... that's pretty similar, from the point of view of the UX. Probably equally risky (see above), but would not need to include a soon-to-be-removed flow in oidc-client-ts.

Comment (jsedanoj) [reply README.md:27]: > I am fine to have this flow even so it is not part of the "OAuth 2.1 draft". Because it fits in this library quiet nicely and its better to have it inside than having multiple implementations outside...
> 
> But please update the README.md description, such that is more clear for the users when to use this flow and when not.

I am waiting for the other branch of the discussion to resolve, because if there are better alternatives... So this message is just to say that if we decide to proceed, I'll improve the README and add the other small improvements.

Comment (longsleep) [reply README.md:27]: > Wouldn't this be equally risky? If the Client is compromised (through for example an XSS attack), the iframe is compromised as well... No?

The security depends on the security headers and the app loaded in the corresponding frame. The security headers can be different for an iframe. So it could be equal if both the app and the IdP have the same security level.



> Do we have an example of this, anyway? I control the IdP (based on KeyCloak).

I am not sure what example you exactly mean. As far as i understand you want to avoid a top level redirect. This is usally solved by using a popup instead (similar to what a native app would do). With a bit of work on your end you can also adapt this to use a visible iframe or a "in app modal dialog" to render an interactive authorize call.

Generally one would do it like this:
1. try silent redirect based authentication (https://authts.github.io/oidc-client-ts/classes/UserManager.html#signinSilent) -> if success: done
2. if fail, check error and do interactive authentication if its a known OpenID connect error
2a. via redirect (https://authts.github.io/oidc-client-ts/classes/UserManager.html#signinRedirect)
2b. via popup (https://authts.github.io/oidc-client-ts/classes/UserManager.html#signinPopup)

The 2a. can also be done inside a different view of your app which the app itself renders into an iframe (login iframe). I don't know any open source example for this, but its not very hard either. That special app mode basically just needs to eventually result in being signed in into the IdP and let the "outer" app know that it shall try silent renew again. In theory something like that could be added to oidc-client-ts as `signinFrame` as it is not much different technically from the signinPopup (https://github.com/authts/oidc-client-ts/blob/8b5a9e7/src/navigators/PopupWindow.ts#L14) but ususally the top level redirect is preferred to allow password managers and similar things to work properly.

So overall, in your case I suggest you think again if you really need to avoid the top level redirect (it will only happen if the user is not signed into the IdP already if you follow the signinSilent first approach).


And last not least, back to the actual topic of this pull request. A use case for the password credential flow is imo some application which is not already a web browser (like a commandline tool etc.) where you need to sign in and want to avoid the sometimes cumbersome break out to a web browser and pasting the callback URL to the app or some such.

Comment (jsedano-emobg) [reply README.md:27]: I think I am starting to understand the misunderstanding.

You are comparing Federated Authenticatoin using Authorization Code Flow vs Resource Owner Password Credentials. Obviously, I agree that in this case you must always go for ACF.

But my comparison is:
* Resource Onwer Password Credentials using KeyCloak as backend.
vs
* Classical user/password form authentication validating the user/password myself in the backend. + implementing the account management. + implementing the password expiration policy. + implementing the refreshing of the access_token. + signing the access_token/id_token. + checking the strengths of passwords. + implementing a backoffice for users management. + Implementing the backoffice for revokation of access_token/id_token. Plus a lot more things that are for free if I use a standard IdP as authenticator for the ROPCF. And gaining nothing, because even if I do it perfectly (which I won't), it will still have the risks of the ROPCF.

(my fault; although this was very clear in my head, probably I have not made it so clear in the PR)

This is why I understand the reason not to want this PR in oidc-client-ts... because... well... it is not OIDC. It just happens to share 99% of the code with an OIDC client implementation, but is not OIDC. Actually, probably it will not even be OAuth2.1, because ROPCF is removed from the draft. This is why I propose to maybe just skip this PR and just publish some inner classes so any third party can do this on their own without duplicating code.

But I am new to this project, I do not understand the politics of oidc-client-ts, I do not know who is the main maintainer or who has the last say, because I see both @pamapa and @longsleep very active, and you guys do not seem to agree on the next step.

Comment (jsedano-emobg) [reply README.md:27]: Oh, and really, there is no offense in rejecting the PR, if there is a good reason to do it. And especially if we can agree on a workaround. You guys are offering the project to humankind, the last thing others can do is be angry about that!

Comment (jsedanoj) [reply README.md:27]: Sorry, I used the work account for the last comment. It is legit: jsedanoj and jsedano-emobg are the same person ;-)

Comment (pamapa) [reply README.md:27]: > I do not know who is the main maintainer
This library has multiple maintainers/reviewers: @kherock, @longsleep, @PascalLuginbuehl and myself. I am trying to find consensus first before we merge this.

We now have two options now:
A: merge this, even if its not part of OAuth2.1 draft (fits in perfect)
B: modify TokenClient, such that it can be used outside (very easy)

Comment (pamapa) [reply README.md:27]: I vote for A, but i am also pretty fine with B!

Comment (longsleep) [reply README.md:27]: I think we can easily do both with little harm done. My concern was more about people using it when they really shouldn't as it is "easier".

Comment (jsedanoj) [reply README.md]: I added a much longer explanation in the README sumarizing this conversation. Please review.

File: src/OidcClient.ts (hunk 2 above)
Comment (jsedanoj) [reply src/OidcClient.ts]: Fixed.

File: src/UserManager.ts (hunk 3 above)
Comment (jsedanoj) [reply src/UserManager.ts]: Fixed.

File: src/UserManager.ts (hunk 4 above)
Comment (jsedanoj) [reply src/UserManager.ts]: Fixed.

File: src/UserManager.ts (hunk 5 above)
Comment (jsedanoj) [reply src/UserManager.ts]: Fixed.

File: README.md [hunk 6]
```diff
@@ -82,3 +84,17 @@ We appreciate feedback and contribution to this repo!
 This project is licensed under the Apache-2.0 license. See the
 [LICENSE](https://github.com/authts/oidc-client-ts/blob/main/LICENSE) file for
 more info.
+
+## Security concerns on Resource Owner Password Credentials flow
```
pamapa, line 88: Please move that **great** but huge text into `docs/index.md` or create a new file in `docs/` and link directly from

`- [Resource Owner Password Credentials Grant](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.3); however, read the security concerns below before using this flow`
->
`- [Resource Owner Password Credentials Grant](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.3); however, read the [security concerns](link...) before using this flow`

pamapa:
great work, when the readme/docu is in shape i will merge...

README.md
```diff
@@ -33,6 +34,7 @@ Implements the following OAuth 2.0 protocols and supports
 - [Building the Source](#building-the-source)
 - [Contributing](#contributing)
 - [License](#license)
+- [Security concerns on Resource Owner Password Credentials flow](#security-concerns-on-resource-owner-password-credentials-flow)
```
pamapa, line 37: this link makes no longer sense

jsedanoj: Thanks, fixed.

pamapa (approved):

PR #763, opened 2022-11-14
preserve id_token and profile even if no openid is in scope of refresh request

### LINKED ISSUE #754: signinSilent using refresh token clears profile info
I have a question.

1. I call signinPopup and log in.
2. User is stored in localStorage with id_token, access_token, refresh_token as well as decoded user info from id_token stored as a profile object.
3. Call signinSilent.
4. oidc-client-ts exchanges the refresh_token for a new access_token and a new refresh_token.
5. localStorage is updated with a new refresh_token and access_token, but the profile object is now empty _(id_token was unchanged)_.

Is that supposed to happen and am I doing something wrong?

Thank you.

pamapa:
… 

Closes/fixes #754

### Checklist


src/ResponseValidator.ts
```diff
@@ -94,20 +94,18 @@ export class ResponseValidator {
...
-                this._validateIdTokenAttributes(response, state.id_token);
-                logger.debug("ID Token validated");
-            }
-            else {
-                // if there's no id_token on the response, copy over id_token from original request
-                response.id_token = state.id_token;
-                // and decoded part too
-                response.profile = state.profile;
-            }
+        if (response.id_token) {
```
longsleep, line 97: I am wondering if that change goes a bit too far as it now tries to parse id_token even if it is not OpenID connect. Probably an esoteric use case, but maybe it is clearer to only change the reuse id_token case and still require the request to have the scope when actually trying parse the response token.

pamapa: i have adapted the code accordingly. I only check for `response.isOpenId` before `_validateIdTokenAttributes` and `_processClaims`, as we do so in all other places within that file. `_validateIdTokenAttributes` handles the non case directly, where `_processClaims` goes over profile, which is only set when isOpenId was true.

src/ResponseValidator.ts
```diff
@@ -95,20 +95,18 @@ export class ResponseValidator {
...
-                this._validateIdTokenAttributes(response, state.id_token);
-                logger.debug("ID Token validated");
-            }
-            else {
-                // if there's no id_token on the response, copy over id_token from original request
-                response.id_token = state.id_token;
-                // and decoded part too
-                response.profile = state.profile;
-            }
+            this._validateIdTokenAttributes(response, state.id_token);
```
longsleep, line 98: I think this still needs to be wrapped with check if there actually is `response.id_token` to avoid throwing in https://github.com/authts/oidc-client-ts/blob/46fe10b1fe78f10d7764e900f19e98eed704d767/src/ResponseValidator.ts#L263

longsleep (changes requested):

pamapa:
@longsleep would be cool if you can have a look at it again, i have fixed what you requested...

longsleep (approved):
LGTM now :+1:

PR #778, opened 2022-11-23
Make it possible to include credentials on all requests.

Sojborg:
When using a proxy or cdn like AWS cloudfront you might need to pass cookies when doing http requests from your web application to your backend service. Ex if your infrastructure requires a token cookie to pass requests through to the backend endpoints.

Right now the oidc-client-ts library doesn't support using credentials on all fetch calls. Because of this the library doesn't support the above use case.

I looked at this pull request: [https://github.com/authts/oidc-client-ts/pull/659](https://github.com/authts/oidc-client-ts/pull/659) which adds support of credentials for the refreshToken fetch request. But it is only supports refreshToken.

In my pull request I changed this to be a fetchRequestCredentials which will put credentials on all fetch calls within the library because in the above use case credentials in needed in all fetch requests.

I could also add support credentials for each fetch request like the refreshToken so there would be settings like this:
- userInfoCredentials
- refreshTokenCredentials
- tokenCredentials
- openConfigurationCredentials
- etc

What is your feedback on this?

### Checklist

### DIFF
--- a/dist/esm/package.json
+++ b/dist/esm/package.json
@@ -1,3 +1,0 @@
-{
-  "type": "module"
-}
--- a/dist/esm/package.json
+++ b/dist/esm/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}
--- a/src/MetadataService.ts
+++ b/src/MetadataService.ts
@@ -17,6 +17,7 @@ export class MetadataService {
     private _metadataUrl: string;
     private _signingKeys: SigningKey[] | null = null;
     private _metadata: Partial<OidcMetadata> | null = null;
+    private fetchRequestCredentials: RequestCredentials | undefined;
 
     public constructor(private readonly _settings: OidcClientSettingsStore) {
         this._metadataUrl = this._settings.metadataUrl;
@@ -29,6 +30,11 @@ export class MetadataService {
         if (this._settings.metadata) {
             this._logger.debug("using metadata from settings");
             this._metadata = this._settings.metadata;
+        }
+
+        if (this._settings.fetchRequestCredentials) {
+            this._logger.debug("using fetchRequestCredentials from settings");
+            this.fetchRequestCredentials = this._settings.fetchRequestCredentials;
         }
     }
 
@@ -49,7 +55,7 @@ export class MetadataService {
         }
 
         logger.debug("getting metadata from", this._metadataUrl);
-        const metadata = await this._jsonService.getJson(this._metadataUrl);
+        const metadata = await this._jsonService.getJson(this._metadataUrl, { credentials: this.fetchRequestCredentials });
 
         logger.debug("merging remote JSON with seed metadata");
         this._metadata = Object.assign({}, this._settings.metadataSeed, metadata);
--- a/src/MetadataService.ts
+++ b/src/MetadataService.ts
@@ -17,7 +17,7 @@ export class MetadataService {
     private _metadataUrl: string;
     private _signingKeys: SigningKey[] | null = null;
     private _metadata: Partial<OidcMetadata> | null = null;
-    private fetchRequestCredentials: RequestCredentials | undefined;
+    private _fetchRequestCredentials: RequestCredentials | undefined;
 
     public constructor(private readonly _settings: OidcClientSettingsStore) {
         this._metadataUrl = this._settings.metadataUrl;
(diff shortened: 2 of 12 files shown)


pamapa:
thanks for contributing, i will merge when all issues are resolved...

dist/esm/package.json
```diff
@@ -1,3 +0,0 @@
-{
-  "type": "module"
-}
```
pamapa, line 3: please leave this file as is, if you really want to remove it, make a separate MR with this...

src/UserInfoService.ts
```diff
@@ -13,7 +14,8 @@ export class UserInfoService {
     protected readonly _logger = new Logger("UserInfoService");
     private readonly _jsonService: JsonService;
 
-    public constructor(private readonly _metadataService: MetadataService) {
+    public constructor(private readonly _metadataService: MetadataService,
+        private readonly _settings: OidcClientSettingsStore) {
```
pamapa, line 18: cosmetics: please move _settings before _metadataService parameter

src/OidcClientSettings.ts
```diff
@@ -104,14 +104,15 @@ export interface OidcClientSettings {
     extraTokenParams?: Record<string, unknown>;
 
     /**
-     * Credentials used by fetch with the refresh request. (default: "same-origin")
+     * Will check the content type header of the response of the revocation endpoint to match these passed values (default: [])
      */
-    refreshTokenCredentials?: "same-origin" | "include" | "omit";
+    revokeTokenAdditionalContentTypes?: string[];
```
pamapa, line 109: 1. please do not change the sequence, it makes it hard to see the diff
2. the rename refreshTokenCredentials to fetchRequestCredentials is a breaking change, maybe its better to leave this as is or we need to keep it and add the new one and merge them with this file. We can then also deprecate the old one...

Sojborg: Okay, so I keep the refreshTokenCredentials even though i removed all usages of it in the code base? I get the idea I just want to make sure that i understand it correctly.

dist/esm/package.json
Sojborg: Yeah, I think i just deleted the whole dist library because i thought it was auto generated :). Will add it back in.

Sojborg:
Thank you for fast replies. Pushed the fixes. Just have a minor question around refreshTokenCredentials.

src/OidcClientSettings.ts
pamapa: yes, in src/OidcClientSettings.ts we then take the new `fetchRequestCredentials` if defined, else we fall-back to the deprecated `refreshTokenCredentials`

something like this
```
this.fetchRequestCredentials = fetchRequestCredentials ? fetchRequestCredentials
    : refreshTokenCredentials ? refreshTokenCredentials : "same-origin";
```

Nice to have: if both are defined print a warning or throw an Error. Not sure if worth...

Sojborg: Ah yes of cause to still support the refreshToken. I pushed the changes. I think it makes sense with a warn if both properties are set. An error is a bit aggressive as the functionality will still work.

src/MetadataService.ts
```diff
@@ -17,6 +17,7 @@ export class MetadataService {
     private _metadataUrl: string;
     private _signingKeys: SigningKey[] | null = null;
     private _metadata: Partial<OidcMetadata> | null = null;
+    private fetchRequestCredentials: RequestCredentials | undefined;
```
pamapa, line 20: `fetchRequestCredentials` -> `_fetchRequestCredentials` (underscore)

pamapa: and align type and default value to others: 
`private _fetchRequestCredentials: RequestCredentials | null = null;`

Sojborg: Default value should be undefined then the credentials property will skipped if the _fetchRequestCredentials isn't set.

Sojborg: Ex here: https://github.com/authts/oidc-client-ts/pull/778/files#diff-dc78e3cc058d3fd2461eec10c5b75ee1c28394a27aadf3ab8a8fe3294547b9e3R58

Sojborg: Renaming done.

pamapa: so only the rename (fetchRequestCredentials -> _fetchRequestCredentials) here...

pamapa (approved):

Sojborg:
And thank you for being fast at replying. Really nice 🥳

Sojborg:
@pamapa when will you make a new npm version with the changes?

pamapa:
today :-)

PR #782, opened 2022-11-25
allow multiple resource parameters

### LINKED ISSUE #769: RFC8707 - Possible Missing Multiple Resources handling for Code Flow Authorization Request
Hello,

My question might be totally naive, but here I go:
We have an SSO from which we're gettings Access/Refresh/Id token in order to later call 2 Services.
We're trying to add the `resource` information as part of the AuthCode Grant during signing in order to generate dedicated token(s ?) to hit both ApiOne and ApiTwo. (see image above from the RFC)
[RFC 8707 - Resource Indicators for OAuth 2.0](https://www.rfc-editor.org/rfc/rfc8707.html)
image

(On refresh the same will apply by sending the last valid refresh token + the appropriate resources)

My question being ... how can this be achieved with the library today for both SignIn, Refresh, etc ...
From the `OidcClient.ts` `resource` seems to be a simple string here:
https://github.com/authts/oidc-client-ts/blob/0e6bfc22ba2181bf52934b1666ac61d9a0a92069/src/OidcClient.ts#L37

and it seems to be forwarded
https://github.com/authts/oidc-client-ts/blob/0e6bfc22ba2181bf52934b1666ac61d9a0a92069/src/OidcClient.ts#L111

And I cannot find the could mapping that to the `&resource=`

pamapa:
Closes/fixes #769

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -76,7 +76,7 @@ export interface CreateSigninRequestArgs {
     // (undocumented)
     request_uri?: string;
     // (undocumented)
-    resource?: string;
+    resource?: string | string[];
     // (undocumented)
     response_mode?: "query" | "fragment";
     // (undocumented)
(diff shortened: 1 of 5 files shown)


tebeco:
Stupid question, does this publish `2.2.1-preview.00x` somewhere by any changes ?

pamapa:
> Stupid question, does this publish `2.2.1-preview.00x` somewhere by any changes ?
No, i simply copy/paste the src directory of this library into a sub-folder of one of my projects and redirect the imports...
Alternate you can create your own local build of this library and then copy paste the dist folder into your node_modules and replace the oidc-client-ts dist temporarily.

longsleep (approved):
LGTM :+1:

PR #798, opened 2022-12-10
Define popup close within 100ms as blocked by the user and throw the relevant error.

alex-eliot:
QoL feature to define the popup closing by a blocker (either as default by browser, or by an add-on/extension) if the popup was closed within 100ms of its creation time.

Currently, whether the popup is closed by a blocker or manually by the user, the same error—"Popup closed by user"—is thrown, making it difficult to decide if the client should fall back to the redirect method in such a case. With this added, the process should be as simple as comparing the thrown error and deciding accordingly.

I am not very familiar with jest, so I didn't know if or how I should write the relevant tests. Please tell me if I should do that. I have only confirmed its functionality via the Parcel sample.

### Checklist

- [x] This PR makes changes to the public API.

### DIFF
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -36,6 +36,12 @@ export class PopupWindow extends AbstractChildWindow {
     public async navigate(params: NavigateParams): Promise<NavigateResponse> {
         this._window?.focus();
 
+        setTimeout(() => {
+            if (!this._window || typeof this._window.closed !== "boolean" || this._window.closed) {
+                this._abort.raise(new Error("Popup blocked by user"));
+            }
+        }, 100);
+
         const popupClosedInterval = setInterval(() => {
             if (!this._window || this._window.closed) {
                 this._abort.raise(new Error("Popup closed by user"));


pamapa:
Can you explain why the existing `const popupClosedInterval = setInterval(() => {...` does not work in this case? What must be done to make it work too?

alex-eliot:
> Can you explain why the existing `const popupClosedInterval = setInterval(() => {...` does not work in this case? What must be done to make it work too?

So I iniitally thought of adding a check inside `popupClosedInterval` that checked if the window was closed in the first iteration (any further wouldn't make sense for an instantly blocked/closed popup). However, the 500ms interval is long enough for a user to manually close it in time, so that may lead to false errors (i.e if the user closes the popup within 500ms of its creation time, the error thrown would be "Popup blocked by user" and we wouldn't want that).

Since a popup blocker either prevents the window from being opened or closes it instantly, the check will only suffice to be in the very first iteration. To make it work inside `popupClosedInterval`, first we'd have to change the interval to 100ms, and throw the error "Popup blocked by user" only if it was closed in the first iteration (meaning that it was closed within 100ms).

pamapa:
see https://github.com/authts/oidc-client-ts/pull/1146

pamapa:
done via #1134

PR #812, opened 2022-12-29
Fix #695: Add filter list to scopes used with refresh tokens

### LINKED ISSUE #695: AzureAD requesting token refresh with all received scopes from signin can be problematic
I'm using oidc-client-ts for an application that logs users in using AzureAD. My signin request to Azure uses their special '.default' [scope](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#request-the-permissions-in-the-app-registration-portal) which provides all of the resource-specific scopes for the resource I'm requesting a token for.

One quirk related to using this scope is that when you make a request that uses it, it must be the _only_ scope in the token request, however the token response will list the `.deafult` scope _along with_ all scopes that it inherently provided.

I.E: 

Request:
```json
{
   ...
   scope: '<client_id>/.default'
}
````

Response:
```json
{
   ...
   scope: '<client_id>/.default <client_id>/User.Read <cliet_id>/User.Calendar'
}
````

This behavior itself is not an issue, however when trying to request a new token using the refresh token via oidc-client-ts the scope from the response is included in the refresh token request and this causees the refresh token request to fail from Azure because the request includes '.default' along with the other scopes.

My current workaround is to add an event listener for user load and overwrite the user's scopes with the scopes used in my initial request. I.E:

```javascript
userManager.events.addUserLoaded((user) => {
    user.scope = '<client_id>/.default'
    userManager.storeUser(currentUser)
});
```
This works but feels a bit clunky and feels like something that may be able to be handled better by the library.

I believe this behavior was added to oidc-client-ts [here](https://github.com/authts/oidc-client-ts/pull/365).

Some potential solutions I've pondered:

1. Providing some way to modify that behavior to use scopes from the initial signin request rather than scopes from the token response
2. Providing some way to explicitly define the scopes used for refresh requests

timgrohmann:
Closes/fixes #695. See discussion in this issue.

### Checklist

### DIFF
--- a/src/OidcClientSettings.ts
+++ b/src/OidcClientSettings.ts
@@ -122,7 +122,7 @@ export interface OidcClientSettings {
     /**
      * Only scopes in this list will be passed in the token refresh request.
      */
-    refreshTokenAllowableScopes?: string | undefined;
+    refreshTokenAllowedScope?: string | undefined;
 }
 
 /**
@@ -173,7 +173,7 @@ export class OidcClientSettingsStore {
 
     public readonly revokeTokenAdditionalContentTypes?: string[];
     public readonly fetchRequestCredentials: RequestCredentials;
-    public readonly refreshTokenAllowableScopes: string | undefined;
+    public readonly refreshTokenAllowedScope: string | undefined;
 
     public constructor({
         // metadata related
@@ -196,7 +196,7 @@ export class OidcClientSettingsStore {
         refreshTokenCredentials,
         revokeTokenAdditionalContentTypes,
         fetchRequestCredentials,
-        refreshTokenAllowableScopes,
+        refreshTokenAllowedScope,
         // extra query params
         extraQueryParams = {},
         extraTokenParams = {},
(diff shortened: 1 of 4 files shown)


src/OidcClientSettings.ts
```diff
@@ -168,6 +173,7 @@ export class OidcClientSettingsStore {
 
     public readonly revokeTokenAdditionalContentTypes?: string[];
     public readonly fetchRequestCredentials: RequestCredentials;
+    public readonly refreshTokenAllowableScopes: string | undefined;
```
pamapa, line 176: Please use type `?: string[];`

src/OidcClientSettings.ts
```diff
@@ -118,6 +118,11 @@ export interface OidcClientSettings {
      * Use this if you need to send cookies to the OIDC/OAuth2 provider or if you are using a proxy that requires cookies
      */
     fetchRequestCredentials?: RequestCredentials;
+
+    /**
+     * Only scopes in this list will be passed in the token refresh request.
+     */
+    refreshTokenAllowableScopes?: string | undefined;
```
pamapa, line 125: Please use type `?: string[];`

src/OidcClientSettings.ts
pamapa: Ah i see you made it the same as `scope?: string`, i guess, makes sense to keep it like you have

pamapa, line 176: I am think about a cosmetic change, maybe `refreshTokenAllowedScope` is a better name?

- without s at the end to match `scope?: string;`
- Allowable vs Allowed

timgrohmann: Yes, I see why `refreshTokenAllowedScope` makes the naming more consistent. Just pushed the new name.

pamapa (approved):

PR #828, opened 2023-01-13
fix(#780): filterProtocolClaims deletes properties required by the IdTokenClaims type

### LINKED ISSUE #780: filterProtocolClaims deletes properties required by the IdTokenClaims type
If filterProtocolClaims is set to true (the default), the properties `iss`, `aud`, `exp` and `iat` that are **_required_** by the IdTokenClaims type are deleted and undefined, thus breaking the defined types.

```typescript
  // ... other options omitted
  filterProtocolClaims: true, // defaults to true
  // ...
});
const user = await userManager.getUser();
const exp = user.profile.exp; // TypeScript says this is a number
console.log(typeof exp) // Should be "number" but is "undefined"
```

**Relevant parts of the code**

These properties are deleted from the response:
https://github.com/authts/oidc-client-ts/blob/bcfe363c685e8f243b43bc588ecba7495f88ffa9/src/ResponseValidator.ts#L26-L41
https://github.com/authts/oidc-client-ts/blob/bcfe363c685e8f243b43bc588ecba7495f88ffa9/src/ResponseValidator.ts#L225-L235
Some of which are mandatory here:
https://github.com/authts/oidc-client-ts/blob/bcfe363c685e8f243b43bc588ecba7495f88ffa9/src/Claims.ts#L109

Badisi:
Closes/fixes #780

### Checklist

### DIFF
--- a/src/ResponseValidator.ts
+++ b/src/ResponseValidator.ts
@@ -16,6 +16,7 @@ import type { RefreshState } from "./RefreshState";
 import type { JwtClaims, IdTokenClaims } from "./Claims";
 
 /**
+ * Protocol claims that could be removed by default from profile.
  * Derived from the following sets of claims:
  * - {@link https://datatracker.ietf.org/doc/html/rfc7519.html#section-4.1}
  * - {@link https://openid.net/specs/openid-connect-core-1_0.html#IDToken}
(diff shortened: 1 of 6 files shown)


src/ResponseValidator.ts
```diff
@@ -16,30 +16,33 @@ import type { RefreshState } from "./RefreshState";
...
+    "at_hash", // https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken
 ] as const;
 
+/**
+ * Protocol claims that should never be removed from profile.
+ * "sub" is needed internally and others should remain required as per the OIDC specs.
+ *
+ * @internal
+ */
+const ForbiddenProtocolClaims = ["sub", "iss", "aud", "exp", "iat"];
```
pamapa, line 44: Please invert the naming. to something like `InternalRequiredProtocolClaims`.

pamapa:
nice work!

src/ResponseValidator.ts
Badisi: Done

pamapa (approved):

Issue #852, opened 2023-01-24
How to re-request custom claims on a logged in user when they have been updated


contractorwolf:
I have users that get claims added to their account after they upgrade, but the call to userManager.getUser() seems to only contain their cached original claims.  How do I get the User manager to request the newly added claims on the user without forcing them to log out and then login again?  Is there a way to invalidate my current client-side account/claims so that the library will go back to auth to get the latest claims?

Right now it only seems that a claim is updated on .getUser() is if the user logs in again?

pamapa:
related to #846

PR #877, opened 2023-02-07
refresh userinfo data on demand

### LINKED ISSUE #846: Load userinfo without refreshing tokens
Is there a way to fetch data from the userinfo endpoint without having to refresh the tokens? 

AFAIU the "loadUserInfo" options loads data upon login, but there's no way to force a refresh of the information.

marcoreni:
Fixes #846 #852

### Checklist

This PR adds a new parameter "refreshUserInfo" to "loadUser", so that it's possible to refresh (and store) updated userinfo claims for the logged user. `loadUser` will refresh the token, if expired, otherwise it will simply reload userinfo. 
In order for this to be enabled, `loadUserInfo` should be enabled in the settings.

NOTE: if the token is refreshed during the procedure, old claims that "disappeared" will be removed. Otherwise, they will be kept.

It also:
- Adds a new configuration property "legacyMergeClaimsBehavior" (defaulted to `true` for backwards compatibility) that, if disabled, overwrites claims instead of transforming types.
- Exposes a new `getUserInfo` method in OidcClient to perform the raw userinfo data retrieval.

LMKWYT

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -287,12 +287,18 @@ export interface OidcAddressClaim {
 // @public
 export class OidcClient {
     constructor(settings: OidcClientSettings);
+    // Warning: (ae-forgotten-export) The symbol "ClaimsService" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    protected readonly _claimsService: ClaimsService;
     // (undocumented)
     clearStaleState(): Promise<void>;
     // (undocumented)
     createSigninRequest({ state, request, request_uri, request_type, id_token_hint, login_hint, skipUserInfo, nonce, response_type, scope, redirect_uri, prompt, display, max_age, ui_locales, acr_values, resource, response_mode, extraQueryParams, extraTokenParams, }: CreateSigninRequestArgs): Promise<SigninRequest>;
     // (undocumented)
     createSignoutRequest({ state, id_token_hint, request_type, post_logout_redirect_uri, extraQueryParams, }?: CreateSignoutRequestArgs): Promise<SignoutRequest>;
+    // (undocumented)
+    getUserInfo(token: string, profile: IdTokenClaims): Promise<IdTokenClaims>;
     // (undocumented)
     protected readonly _logger: Logger;
     // (undocumented)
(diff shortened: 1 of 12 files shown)


pamapa:
Thanks for providing this. I like the claims parts, yes it makes it easier when we split it in its own service. This code is almost ready to merge. With the `refreshUserInfo` i am not so happy yet because of the edge case handling (expired, signinSilent).

Idea:
- keep  `refreshUserInfo` in getUser, but without edge case handling, e..g  the caller must take care the User is still valid... if not doing `signinSilent`, ...

Simple without edge cases:
```
if (refreshUserInfo) {
    logger.debug("refreshing user info");
    user.profile = await this._client.getUserInfo(user);
    await this.storeUser(user);
    logger.debug("user updated in storage");
}
```

pamapa:
Would be nice if you could split this up into two merge requests.

marcoreni:
> keep refreshUserInfo in getUser, but without edge case handling, e..g the caller must take care the User is still valid... if not doing signinSilent, ...

What you added as an example was exactly my first implementation. After that, my reasoning was: UserManager is described as `higher level API`, so it should help users to obtain what they want without the hassle of multiple checks. 

I can definitely simplify this. What do you think should happen if the token expired? We should throw an exception?

> Would be nice if you could split this up into two merge requests.

Yeah no problem, you mean one for "splitting the claims service and changing the merge behavior" and the other for "adding the userinfo refresh behavior" ?

mattsGitHub123:
@marcoreni Curious if your still working on this? I've got a use case for this functionality and would love to be able to use it.

tbm206:
> @marcoreni Curious if your still working on this? I've got a use case for this functionality and would love to be able to use it.

Same here

marcoreni:
Hi all,

sadly, we did not have time to wait on the discussion on https://github.com/authts/oidc-client-ts/pull/881, so we forked the library and implemented a basic version of the "on demand update" (that served our purposes). The implementation is available on [these](https://github.com/MDS-BD/oidc-client-ts/commit/d325548db9ef2301aacd7f5c7b9e6307893dd1ee) [two](https://github.com/MDS-BD/oidc-client-ts/commit/a34dfcf436fd09f932b6b3582a74f6ec4a36b659) commits but I think it's not ready for merge for the following reasons:
1- it's based on https://github.com/authts/oidc-client-ts/pull/881 which has been superseded by https://github.com/authts/oidc-client-ts/pull/904
2- it does not have tests
3- it does not integrate the changes requested above

I don't think I'll have time to tackle this again soon.

PR #881, opened 2023-02-09
feat: new claimsService and new merge behavior

marcoreni:
Closes/fixes

### Checklist

First chunk of #877 splitting. This PR introduces the new merge behavior and moves all the claims related logic inside a dedicated service


src/OidcClientSettings.ts
```diff
@@ -92,6 +92,11 @@ export interface OidcClientSettings {
      * Otherwise, they are added to an array as distinct objects for the claim type. (default: false)
      */
     mergeClaims?: boolean;
+    /**
+     * Indicates if the library should use the legacy merge behavior, which mutates string claims into arrays whenever the data is updated on the remote.
+     * This behavior is enabled by default on v2 and will be removed on v3, since it's not a deterministic way of handling claims.
+     */
```
pamapa, line 98: - add  `(default: true)` at the end, like all other props have
- rename `legacyMergeClaimsBehavior` to `mergeClaimsLegacyBehavior`

src/OidcClientSettings.ts
```diff
@@ -168,6 +173,10 @@ export class OidcClientSettingsStore {
     public readonly clockSkewInSeconds: number;
     public readonly userInfoJwtIssuer: "ANY" | "OP" | string;
     public readonly mergeClaims: boolean;
+    /**
+     * TODO: remove me on v3
+     */
+    public readonly legacyMergeClaimsBehavior: boolean;
```
pamapa, line 179: please remove this TODO comment

src/ClaimsService.ts
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        return result;
+    }
+
+    public mergeClaims(claims1: UserProfile, claims2: JwtClaims): UserProfile;
+    public mergeClaims(claims1: JwtClaims, claims2: JwtClaims): JwtClaims;
+    public mergeClaims(claims1: JwtClaims | UserProfile, claims2: JwtClaims): JwtClaims | UserProfile {
+        // TODO: remove on next major version
```
pamapa, line 97: please remove the TODO comment

src/ClaimsService.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (this._settings.legacyMergeClaimsBehavior) {
+            return this.legacyMergeClaims(claims1 as UserProfile, claims2);
+        }
+
+        const result = { ...claims1 };
+
```
pamapa, line 103: this newline is not needed

src/OidcClientSettings.ts
```diff
@@ -92,6 +92,11 @@ export interface OidcClientSettings {
      * Otherwise, they are added to an array as distinct objects for the claim type. (default: false)
      */
     mergeClaims?: boolean;
+    /**
+     * Indicates if the library should use the legacy merge behavior, which mutates string claims into arrays whenever the data is updated on the remote.
+     * This behavior is enabled by default on v2 and will be removed on v3, since it's not a deterministic way of handling claims.
```
pamapa, line 97: Please remove "This behavior is enabled by default on v2 and will be removed on v3, since it's not a deterministic way of handling claims."

Maybe we only change the default from `true` to `false` and deprecate it, lets see later...

pamapa:
Nice work! Only some small things to fix...

src/OidcClientSettings.ts
marcoreni: Fixed

src/OidcClientSettings.ts
marcoreni: Fixed

src/ClaimsService.ts
marcoreni: Fixed

src/ClaimsService.ts
marcoreni: Fixed

src/OidcClientSettings.ts
marcoreni: I did already remove this, let me know if you like to add some different message :)

pamapa: For me the legacy behavior looks like a bug!

This merge request fixes that bug. For users who would require the old behavior (i doubt it for now) we provide this option to switch back. In that case the default is `false` and we can already mark it as deprecated.
Like this, we can also will find out sooner if somebody actually requires that strange behavior and for what reason...

I can reflect this by a minor version bump and a special note in the release notes.

src/OidcClientSettings.ts
```diff
@@ -92,6 +92,8 @@ export interface OidcClientSettings {
      * Otherwise, they are added to an array as distinct objects for the claim type. (default: false)
      */
     mergeClaims?: boolean;
+    /** Indicates if the library should use the legacy merge behavior, which mutates string claims into arrays whenever the data is updated on the remote. (default: false) */
+    mergeClaimsLegacyBehavior?: boolean;
```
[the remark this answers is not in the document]
pamapa: funny: we introduce an option and make it directly deprecated, but otherwise we can not get rid of it any time soon...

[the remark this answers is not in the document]
pamapa: @Badisi Do you see any use case for the broken old legacy merging?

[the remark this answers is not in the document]
Badisi: This whole "merge claims mess" was introduced back in 2019 with this PR [#754](https://github.com/IdentityModel/oidc-client-js/pull/754) but existed before that.
Complains followed [(#861)](https://github.com/IdentityModel/oidc-client-js/issues/861) after the PR and the `mergeClaims: boolean` option was introduced to mitigate that.

I have no idea why this behavior of keeping the old and new values into arrays was made, and I don't know any use cases that would required that.. Maybe @brockallen could enlighten us here ? But my understanding of `mergeClaims()` is that it is used to merge the `IdToken` with the response of `UserInfo`. So I could see why it was necessary somehow to keep values of keys that would exists in the two but were different.

My concerns:

1. With @marcoreni new proposal, we are making `UserInfo` the master, as we are overriding any existing key of `claims1` with the value of `claims2`. Is it a good assumption ? You tell me.

2. Also, AFAIU, with this new algorithm we are keeping any keys in `claims1` that do not exist in `claims2` and replacing the existing ones or merging them if there are objects. But for arrays, the logic is different as we are just overriding them, where I would have expected to also keep the existing values and only add new ones or replace existing).
   ``` 
   claims1 { a: 'a', b: [1,2,3], c: { d: 'd' } }
   claims2 { b: [4], c: { e: 'e' } } 
   actual results: { a: 'a', b: [4], c: { d: 'd', e: 'e' } }
   my expectation: { a: 'a', b: [1,2,3,4], c: { d: 'd', e: 'e' } }
   ```

3. Even with `mergeClaims: false` we are still, somehow, merging the two as `claims1` would receive any new keys from `claims2` and got its already existing keys overridden. So as a user perspective it doesn't feel quite right.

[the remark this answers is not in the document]
Badisi: While I'm thinking about it, why are we trying to make this `mergeClaims()` so complex ?

This isn't clear in the RFC, but can we say that the `IdToken` is always a subset of what is returned by `UserInfo` ? And if that's true, then `UserInfo` is always the up-to-date reference and a simple merge like `{ ...claims1, ..claims2 }` should be enough ?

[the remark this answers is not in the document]
marcoreni: @Badisi this PR is part of a larger PR (#877) that aims at adding the capability to refresh UserInfo without refreshing the token.

So, even if UserInfo should be a superset of IdToken (I was also trying to find out some spec about this), if we add the refresh of UserInfo, in order to have a clean and definitive solution we may need to address some more complex scenarios (as I mentioned [here](https://github.com/authts/oidc-client-ts/issues/846#issuecomment-1406430179))

My "ideal solution", regarding UserInfo, is that we should have a "core object", that does NOT contain any of the UserInfo data, that is the master of the information. 
If the client has "loadUserInfo" enabled, the userinfo stuff is retrieved and cached separately, and merged "on demand". Whenever the "userinfo" is updated, the userinfo cache is completely overwritten (so, no leftovers).

This, however, would introduce some heavy refactoring on the library, unless we find some workaroung (ie. not decoding the jwt while validating the response and caching only userinfo inside "profile", then merging on retrieval, something like that).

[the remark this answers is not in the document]
pamapa: @Badisi Thanks for the excellent summary!

> But my understanding of `mergeClaims()` is that it is used to merge the `IdToken` with the response of `UserInfo`.

I have the same understanding.

To your concern 1:
It can still be controlled via the settings `loadUserInfo`, which is `false` by default. So no surprise here.
To your concern 2:
I agree with you about the arrays logic, lets combine them when merging. There are anyway not many array fields, most of the claims have string or object as type. This is one i have found, which can be an array: `aud?: string | string[];`
To your concern 3:
I agree with you, the existing behavior should be kept if `mergeClaims` is `false`.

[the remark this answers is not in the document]
pamapa: hmm, when i look at the current test code, it does exactly what @Badisi and myself specified above. Do i miss something?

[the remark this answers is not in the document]
pamapa: new:
```
...
else if (result[claim] !== value) {
  if (this._settings.mergeClaims) {
      if (typeof value === "object") {
          result[claim] = this._mergeClaims(previousValue as UserProfile, value);
      } else {
          result[claim] = value;
      }
  }
  else {
      result[claim] = [previousValue, value];
  }
}
...
```

brockallen:
The intent is to merge the identity claims in the id_token and the userinfo endpoint. And it's difficult to merge those if you have duplicates/conflicts in the sets of values (which unfortunately some token servers do). But those values "merged" all in one place are useful.

The merge used to work fairly well, then at some point someone came along and submitted a PR (of course, for their specific scenario) that actually made things worse for most everyone else. I am pretty sure I reverted it, but I don't recall now, so the state of things might have not been ideal.

pamapa:
At the end it might be better to not merge at all, but store both claims objects within the user object. Something like:
```
profile["id_token"] = {id token claims}
profile["user_info"] = {user info claims}
or
claims_id_token = {id token claims}
claims_user_info =  {user info claims}
or better naming...
```

Like this the user of the library can decide what is the best to choose and we could still calculate an additional `claims` object, which is the merged of both if `mergeClaims: true`. How to merge still needs to be defined first, as it is not trivial.

brockallen:
Well, the original implementation was decent, so you might check the history and compare to what it does now. But what I'd suggest then is to leave how it used to work (pre-PR I mentioned) and then make that extensible so people can override if needed. I still suggest putting it all in one place is important, as it shields the client code from needing to understand the details of claim in the id_token vs claims in the user info endpoint. And that might change as well in the token server based on config, something which the client logic will benefit being shielded from.

pamapa:
> I still suggest putting it all in one place is important, as it shields the client code from needing to understand the details of claim in the id_token vs claims in the user info endpoint.

I came to the same conclusion. Lets keep this simple and stupid. In case of merging the only question is; what is happening with arrays. Currently we are aggregating them, which is not the best solution (e.g. `role: ["admin"]`+`role: ["user"]`). As such we allow the user of this library to choose what shall happen with arrays (aggregate or replace):
- not exists -> add
- array -> aggregate or replace
- other -> replace
- object (not array) -> merge

I am not so sure if the current aggregate array behavior makes sense in the long run...

Badisi:
@pamapa, do not forget about the upcoming `refreshUserInfo()` feature.

We will have 2 use cases:
1. merge after sign-in (where we will receive and merge an up-to-date **IdToken** and **UserInfo**)
2. merge when refreshing user info (where we will received an up-to-date **UserInfo** but merge it in an outdated **UserProfile**)

In case 2, we might have received something in the IdToken at the beginning that is no more true in the new UserInfo (like a role that was removed to the user).

pamapa:
> In case 2, we might have received something in the IdToken at the beginning that is no more true in the new UserInfo (like a role that was removed to the user). So in this case : array -> substract

The idea is work there only with `array replace`, either `aggregate` does not exist or we throw an error. At the end its probably easier to only support `array replace`. In my opinion that should be enough to make `refreshUserInfo` work!

Maybe it makes sense for the `refreshUserInfo` use case to start again with the old IdToken and then merge it with the fresh userInfo, instead of using the already merge object...

marcoreni:
@pamapa I agree that "starting from scratch" while updating userinfo is better, but [this line](https://github.com/authts/oidc-client-ts/blob/de93b53bc9594b8dc897c3c6eb602154b541de4a/src/ResponseValidator.ts#L68) makes me wonder... What happens if isOpenId is false?

pamapa:
> but [this line](https://github.com/authts/oidc-client-ts/blob/de93b53bc9594b8dc897c3c6eb602154b541de4a/src/ResponseValidator.ts#L68) makes me wonder... What happens if isOpenId is false?

if openid is not true, means there is no id_token. Thus userInfo will be the first to start the user.profile

brockallen:
> Thus userInfo will be the first to start the user.profile

Userinfo doesn't exist without OIDC, and thus there is no user profile without OIDC.

marcoreni:
So, trying to recap:
1- "Refresh user info" will create `profile` from scratch, re-decoding id_token claims and merging userinfo claims. The updated data will be stored just like today.
2- merge function, when in `new behavior mode`, will perform the following:
```
a. not exists -> add
b. array -> replace (since userinfo should be a superset of idtoken claims, we should be alright)
c. object (not array) -> merge
d. other -> replace
```
3- merge function will be used in both "signin scenario" and "refreshuserinfo scenario", with the same behavior

Questions:
1- 2c. (object.merge) is currently (as of v2.2.1) controlled by setting `mergeClaims` flag. Are we saying that in the "new world" we want this as the standard behavior and not to be controlled by any flag?

Have I missed anything?

(I'll have to close both this PR and the other one since the desiderata is quite different from what I implemented. I hope to be able to work on this soon.)

Badisi:
1. The current behavior of the `mergeClaims` settings is misleading to me as a merge still occur even when set to **false** (ie. 2a, 2b, 2d). But if `mergeClaims: false` equals `not merge at all`, then getting `UserInfo` would be completely useless, as the info would never go up to the user. So either we remove the option or we rename it to reflect the fact that it is only doing 2c.
But I suspect @pamapa's thumb up to go for complete removal. @pamapa, can you confirm ?

3. I don't think so, grazie for the recap 😉

marcoreni:
So!

The new "merge behavior" was not so different from what was originally implemented in this PR. I
- implemented the merge function discussed
- deprecated `mergeClaims` and not considered it in the new implementation

Hopefully, this is good to go. 

I see there are some conflicts, let me know if you need me to rebase.

pamapa:
please resolve the conflicts otherwise i can not start the unit-tests...

marcoreni:
@pamapa Rebased, squashed and ready to go.

src/OidcClientSettings.ts
```diff
@@ -90,8 +90,14 @@ export interface OidcClientSettings {
     /**
      * Indicates if objects returned from the user info endpoint as claims (e.g. `address`) are merged into the claims from the id token as a single object.
      * Otherwise, they are added to an array as distinct objects for the claim type. (default: false)
+     * @deprecated since version 2.3.0. It's not a deterministic way of handling claims.
      */
     mergeClaims?: boolean;
```
pamapa, line 95: Maybe it makes sense to keep this settings in the long turn in the way it was supposed:

From the description:
```
Indicates if objects returned from the user info endpoint as claims (e.g. `address`) are merged into the claims from the id token as a single object.
Otherwise, they are added to an array as distinct objects for the claim type. (default: false)
```
-> true = merged, false = added/appended as an array.

-> mergeClaims = true -> the new behavior of array handling
-> mergeClaims = false -> the legacy behavior

We keep the default as is and in a major release we change the default to `true`.
The `mergeClaimsLegacyBehavior` is then not needed

What do you think? Does that make sense?

pamapa: If we agree on the above it makes sense to rename `legacyMergeClaims` to `appendClaims` and **not** call it from `mergeClaims`. Instead we make and if/else on the caller side depending on `mergeClaims`.
-> easier to understand and easier to follow the unit-tests.

marcoreni: @pamapa frankly I just want this implementation to go ahead.

The current behavior is wrong, wherever we spin it we will have some breaking changes unless we create new attributes to keep this wrong behavior.

Right now `mergeClaims`, when enabled, works as a `deep merging` mechanism, which we will have by default in the new implementation, and changing its meaning will break some usage somewhere.

Whichever solution makes it clearer for you makes sense to me, I'll stick to the new implementation and try to forget about all of this ASAP, but we have 40 comments on this PR and I think we need to land it somehow.

pamapa: @Badisi Should we keep the setting `mergeClaims` in the long run?

pamapa: > I'll stick to the new implementation and try to forget about all of this ASAP

I will help you here. First step see #904

Badisi: I agree with @marcoreni that this whole `mergeClaims` thing is wrong from the start and that it's taking us too much discussion. TBH I don't even recall why we were talking about rewriting it at the first place. I thought that @marcoreni requirement was only to be able to reload UserInfo ?

So do we have any issue today with this merge mechanism ? If not then I would suggest to leave it as is, as it seems to work well for everybody (else we would have had other complains). And if we really need to refactor it I would go with a `mergeClaimsStrategy=['none','merge','replace']` (or something alike), not so disruptive and with all use cases covered.

pamapa: > requirement was only to be able to reload UserInfo

will not work with the current wrong merge behavior...

Badisi: I'm confused..
- The first time, `IdToken` will be merged with `UserInfo` data.
- The second time (when asking to refresh user info), we will use the current `IdToken` and merge it with the newly received `UserInfo` data.

This should be exactly the same.. why it wouldn't work ?

pamapa: @Badisi You are right lets stop this merge request for now. And implement the feature like you said.

Merging claims is not perfect, we can change that later...

pamapa:
@marcoreni Are you still interested to land this? I am not planing a 3.0.0, as such we can make breaking changes, lets focus your engineered algorithm of merging and drop the current merging + drop the mergingClaims settigs....

Issue #895, opened 2023-02-16
Feature request make pkce optional


satanshiro:
I stumbled into a problem where I need to disable PKCE validation.
why? 
lets say I have one frontend app that does the login process but I would like to send a redirect url to another app in my organization that will finish the login process and get a code. (considering that this external redirect url is a valid one by the idp).
So I either have to start the login process from the second app or disable pkce.
one might say just duplicate the login component to every app in the organization but I'm not sure that's effective considering maintenance costs and different apps might be different languages/technologies/deployment behavior altogether.
I am specifically using keycloak and in their documentation if the initiating client does not send a code challenge, pkce validation will be disabled but oidc-ts does not give me the option to disable it
refference:
![image](https://user-images.githubusercontent.com/38865738/219323205-cf5ad6e6-2ef6-486f-bf2c-89e5c95f30a1.png)
I have tested locally changing the code and would like to open a pr if people agree.
have a nice day

pamapa:
I would accept such a merge request if the changes are not too invasive. Best would be if you try this in a local copy of the library and propose changes once you have something working...

satanshiro:
yes I tested locally and it works, I can't push a branch because I don't have permissions, but these are the main changes:
![image](https://user-images.githubusercontent.com/38865738/219365458-576ccb4f-4e92-4e59-9245-87e9f3aff926.png)
in oidcclientsettings:
![image](https://user-images.githubusercontent.com/38865738/219365550-014493d3-fbd4-4ee3-adbd-0c51fd038e71.png)
![image](https://user-images.githubusercontent.com/38865738/219365681-2b7dca1c-dc4b-4042-952a-d82b72ff3bc3.png)

pamapa:
please create a merge request. You fork this project and create a branch on the forked project. On this branch you make a commit with your changes. Within this project you create via UI a pull request. Then i can review the changes and give you feedback...

First feedback: in oidcclientsettings: please a a proper type doc comment, which explains that specific feature, such that others understand when and why its useful and it should end with `(default: false)`

pamapa:
It might be better to handle this in `SigninState` if `code_verifier` is not set `code_challenge` is not set and your if above works too. You can pass the new option `disablePKCE` through the code like `extraQueryParams`...

satanshiro:
feature merged waiting for release

satanshiro:
@pamapa i saw that last release was a month ago, is there a plan for release soon?

pamapa:
[...]

PR #897, opened 2023-02-16
feature to toggle off pkce

satanshiro:
Closes/fixes #issue
https://github.com/authts/oidc-client-ts/issues/895
### Checklist

### DIFF
--- a/src/OidcClientSettings.ts
+++ b/src/OidcClientSettings.ts
@@ -116,6 +116,9 @@ export interface OidcClientSettings {
      * Will check the content type header of the response of the revocation endpoint to match these passed values (default: [])
      */
     revokeTokenAdditionalContentTypes?: string[];
+    /**
+     * Will disable pkce validation, changing to true will not append to sign in request code_challenge and code_challenge_method. (default: false)
+     */
     disablePKCE?: boolean;
     /**
      * Sets the credentials for fetch requests. (default: "same-origin")
--- a/src/OidcClientSettings.ts
+++ b/src/OidcClientSettings.ts
@@ -182,6 +182,7 @@ export class OidcClientSettingsStore {
     public readonly fetchRequestCredentials: RequestCredentials;
     public readonly refreshTokenAllowedScope: string | undefined;
     public readonly disablePKCE: boolean;
+    
     public constructor({
         // metadata related
         authority, metadataUrl, metadata, signingKeys, metadataSeed,
--- a/src/OidcClientSettings.ts
+++ b/src/OidcClientSettings.ts
@@ -116,7 +116,7 @@ export interface OidcClientSettings {
      * Will check the content type header of the response of the revocation endpoint to match these passed values (default: [])
      */
     revokeTokenAdditionalContentTypes?: string[];
-
+    disablePKCE?: boolean;
     /**
      * Sets the credentials for fetch requests. (default: "same-origin")
      * Use this if you need to send cookies to the OIDC/OAuth2 provider or if you are using a proxy that requires cookies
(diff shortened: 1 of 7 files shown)


src/ResponseValidator.ts
```diff
@@ -171,9 +171,6 @@ export class ResponseValidator {
             logger.throw(new Error("Expected code in response"));
         }
```
satanshiro, line 173: here I was not sure what to do here I did not see a reason to save in state the pkcedisabled option,
instead if code verifier does not exist on client I do not check it. I can say for sure that a pkce enabled initiated login will not succeed at server side if you don't send the code, so I think this is safe to remove let me know if you think I'm wring.

src/TokenClient.ts
```diff
@@ -86,9 +86,6 @@ export class TokenClient {
         if (!args.code) {
             logger.throw(new Error("A code is required"));
         }
```
satanshiro, line 88: if you look here 
image
it seemed someone wanted to make it optional but didn't go through it
instead threw  an error here.

src/ResponseValidator.ts
```diff
@@ -171,9 +171,6 @@ export class ResponseValidator {
...
 
-        if (!state.code_verifier && response.code) {
-            logger.throw(new Error("Unexpected code in response"));
```
satanshiro, line 175: here I was not sure what to do here I did not see a reason to save in state the pkcedisabled option,
instead if code verifier does not exist on client I do not check it. I can say for sure that a pkce enabled initiated login will not succeed at server side if you don't send the code, so I think this is safe to remove let me know if you think I'm wring.

src/TokenClient.ts
```diff
@@ -86,9 +86,6 @@ export class TokenClient {
...
-        if (!args.code_verifier) {
-            logger.throw(new Error("A code_verifier is required"));
```
satanshiro, line 90: if you look here 
image
it seemed someone wanted to make it optional but didn't go through it
instead threw  an error here.

src/SigninRequest.ts
```diff
@@ -87,7 +87,7 @@ export class SigninRequest {
         this.state = new SigninState({
             data: state_data,
             request_type,
-            code_verifier: true,
+            code_verifier: !optionalParams.disablePKCE,
```
pamapa, line 90: remove disablePKCE from optionalParams otherwise its lands in params

src/SigninRequest.ts
```diff
@@ -104,7 +104,7 @@ export class SigninRequest {
         }
 
         parsedUrl.searchParams.append("state", this.state.id);
-        if (this.state.code_challenge) {
+        if (this.state.code_challenge && !optionalParams.disablePKCE) {
```
pamapa, line 107: `&& !optionalParams.disablePKCE` should not be needed, as `this.state.code_challenge` should be undefined, because of adapted `SigninState` above

src/ResponseValidator.ts
pamapa: i expect in your case `state.code_verifier=false` and in the response no code. The if should still working...

src/OidcClientSettings.ts
```diff
@@ -178,7 +181,7 @@ export class OidcClientSettingsStore {
     public readonly revokeTokenAdditionalContentTypes?: string[];
     public readonly fetchRequestCredentials: RequestCredentials;
     public readonly refreshTokenAllowedScope: string | undefined;
-
```
pamapa, line 181: please keep a newline between members and constructor

src/ResponseValidator.ts
satanshiro: are you sure? what happens is code_verifier = false, but code does come back that's the whole point I can exchange code for a token without pkce.

pamapa (approved):

pamapa:
thanks for contributing

PR #909, opened 2023-03-04
use web crypto instead of cryptojs

### LINKED ISSUE #275: replace crypto-js
When using `oidc-client-ts` in an Angular application, the following warning is raised:

> Warning: node_modules/oidc-client-ts/dist/esm/oidc-client-ts.js depends on 'crypto-js/enc-base64.js'. CommonJS or AMD dependencies can cause optimization bailouts.
> For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies

There is a solution to explicitely make the warning silent in the application :

```json
"architect": {
    "build": {
        "options": {
            "allowedCommonJsDependencies": [
                "crypto-js"
            ]
        }
    }
}
```

But I was thinking that, maybe, `oidc-client-ts` could simply migrates to [crypto-es](https://github.com/entronad/crypto-es). What do you think ?

Because unfortunately `crypto-js` doesn't seem to care about moving to es module..  [#60](https://github.com/brix/crypto-js/issues/60) and [#236](https://github.com/brix/crypto-js/pull/236)

43081j:
Closes/fixes #275

### Checklist

- [ ] replace test runner/library
- [ ] update docs for interface change

### DIFF
--- a/src/SigninRequest.ts
+++ b/src/SigninRequest.ts
@@ -8,7 +8,7 @@ import { SigninState } from "./SigninState";
  * @public
  * @see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
  */
-export interface SigninRequestArgs {
+export interface SigninRequestCreateArgs {
     // mandatory
     url: string;
     authority: string;
@@ -49,12 +49,20 @@ export interface SigninRequestArgs {
  * @public
  */
 export class SigninRequest {
-    private readonly _logger = new Logger("SigninRequest");
+    private static readonly _logger = new Logger("SigninRequest");
 
     public readonly url: string;
     public readonly state: SigninState;
 
-    public constructor({
+    private constructor(args: {
+        url: string;
+        state: SigninState;
+    }) {
+        this.url = args.url;
+        this.state = args.state;
+    }
+
+    public static async create({
         // mandatory
         url, authority, client_id, redirect_uri, response_type, scope,
         // optional
@@ -65,33 +73,33 @@ export class SigninRequest {
         extraTokenParams,
         disablePKCE,
         ...optionalParams
-    }: SigninRequestArgs) {
+    }: SigninRequestCreateArgs): Promise<SigninRequest> {
         if (!url) {
-            this._logger.error("ctor: No url passed");
+            this._logger.error("create: No url passed");
             throw new Error("url");
         }
         if (!client_id) {
-            this._logger.error("ctor: No client_id passed");
+            this._logger.error("create: No client_id passed");
             throw new Error("client_id");
         }
         if (!redirect_uri) {
-            this._logger.error("ctor: No redirect_uri passed");
+            this._logger.error("create: No redirect_uri passed");
             throw new Error("redirect_uri");
         }
         if (!response_type) {
-            this._logger.error("ctor: No response_type passed");
+            this._logger.error("create: No response_type passed");
             throw new Error("response_type");
         }
         if (!scope) {
-            this._logger.error("ctor: No scope passed");
+            this._logger.error("create: No scope passed");
             throw new Error("scope");
         }
         if (!authority) {
-            this._logger.error("ctor: No authority passed");
+            this._logger.error("create: No authority passed");
             throw new Error("authority");
         }
 
-        this.state = new SigninState({
+        const state = await SigninState.create({
             data: state_data,
             request_type,
             url_state,
@@ -111,13 +119,13 @@ export class SigninRequest {
             parsedUrl.searchParams.append("nonce", nonce);
         }
 
-        let state = this.state.id;
+        let stateParam = state.id;
         if (url_state) {
-            state = `${state}${URL_STATE_DELIMITER}${url_state}`;
+            stateParam = `${stateParam}${URL_STATE_DELIMITER}${url_state}`;
         }
-        parsedUrl.searchParams.append("state", state);
-        if (this.state.code_challenge) {
-            parsedUrl.searchParams.append("code_challenge", this.state.code_challenge);
+        parsedUrl.searchParams.append("state", stateParam);
+        if (state.code_challenge) {
+            parsedUrl.searchParams.append("code_challenge", state.code_challenge);
             parsedUrl.searchParams.append("code_challenge_method", "S256");
         }
 
@@ -134,6 +142,9 @@ export class SigninRequest {
             }
         }
 
-        this.url = parsedUrl.href;
+        return new SigninRequest({
+            url: parsedUrl.href,
+            state,
+        });
     }
 }
--- a/src/SigninState.ts
+++ b/src/SigninState.ts
@@ -4,6 +4,31 @@
 import { Logger, CryptoUtils } from "./utils";
 import { State } from "./State";
 
+/** @public */
+export interface SigninStateArgs {
+    id?: string;
+    data?: unknown;
+    created?: number;
+    request_type?: string;
+
+    code_verifier?: string;
+    code_challenge?: string;
+    authority: string;
+    client_id: string;
+    redirect_uri: string;
+    scope: string;
+    client_secret?: string;
+    extraTokenParams?: Record<string, unknown>;
+    response_mode?: "query" | "fragment";
+    skipUserInfo?: boolean;
+    url_state?: string;
+}
+
+/** @public */
+export type SigninStateCreateArgs = Omit<SigninStateArgs, "code_verifier"> & {
+    code_verifier?: string | boolean;
+};
+
 /**
  * @public
  */
@@ -32,36 +57,11 @@ export class SigninState extends State {
 
     public readonly skipUserInfo: boolean | undefined;
 
-    public constructor(args: {
-        id?: string;
-        data?: unknown;
-        created?: number;
-        request_type?: string;
-        url_state?: string;
-
-        code_verifier?: string | boolean;
-        authority: string;
-        client_id: string;
-        redirect_uri: string;
-        scope: string;
-        client_secret?: string;
-        extraTokenParams?: Record<string, unknown>;
-        response_mode?: "query" | "fragment";
-        skipUserInfo?: boolean;
-    }) {
+    private constructor(args: SigninStateArgs) {
         super(args);
 
-        if (args.code_verifier === true) {
-            this.code_verifier = CryptoUtils.generateCodeVerifier();
-        }
-        else if (args.code_verifier) {
-            this.code_verifier = args.code_verifier;
-        }
-
-        if (this.code_verifier) {
-            this.code_challenge = CryptoUtils.generateCodeChallenge(this.code_verifier);
-        }
-
+        this.code_verifier = args.code_verifier;
+        this.code_challenge = args.code_challenge;
         this.authority = args.authority;
         this.client_id = args.client_id;
         this.redirect_uri = args.redirect_uri;
@@ -71,6 +71,17 @@ export class SigninState extends State {
 
         this.response_mode = args.response_mode;
         this.skipUserInfo = args.skipUserInfo;
+    }
+
+    public static async create(args: SigninStateCreateArgs): Promise<SigninState> {
+        const code_verifier = args.code_verifier === true ? CryptoUtils.generateCodeVerifier() : (args.code_verifier || undefined);
+        const code_challenge = code_verifier ? (await CryptoUtils.generateCodeChallenge(code_verifier)) : undefined;
+
+        return new SigninState({
+            ...args,
+            code_verifier,
+            code_challenge,
+        });
     }
 
     public toStorageString(): string {
@@ -94,9 +105,9 @@ export class SigninState extends State {
         });
     }
 
-    public static fromStorageString(storageString: string): SigninState {
+    public static fromStorageString(storageString: string): Promise<SigninState> {
         Logger.createStatic("SigninState", "fromStorageString");
         const data = JSON.parse(storageString);
-        return new SigninState(data);
+        return SigninState.create(data);
     }
 }
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -46,7 +46,7 @@ export class CheckSessionIFrame {
 }
 
 // @public (undocumented)
-export interface CreateSigninRequestArgs extends Omit<SigninRequestArgs, "url" | "authority" | "client_id" | "redirect_uri" | "response_type" | "scope" | "state_data"> {
+export interface CreateSigninRequestArgs extends Omit<SigninRequestCreateArgs, "url" | "authority" | "client_id" | "redirect_uri" | "response_type" | "scope" | "state_data"> {
     // (undocumented)
     redirect_uri?: string;
     // (undocumented)
@@ -622,7 +622,8 @@ export type SigninRedirectArgs = RedirectParams & ExtraSigninRequestArgs;
 
 // @public (undocumented)
 export class SigninRequest {
-    constructor({ url, authority, client_id, redirect_uri, response_type, scope, state_data, response_mode, request_type, client_secret, nonce, url_state, resource, skipUserInfo, extraQueryParams, extraTokenParams, disablePKCE, ...optionalParams }: SigninRequestArgs);
+    // (undocumented)
+    static create({ url, authority, client_id, redirect_uri, response_type, scope, state_data, response_mode, request_type, client_secret, nonce, url_state, resource, skipUserInfo, extraQueryParams, extraTokenParams, disablePKCE, ...optionalParams }: SigninRequestCreateArgs): Promise<SigninRequest>;
     // (undocumented)
     readonly state: SigninState;
     // (undocumented)
@@ -630,7 +631,7 @@ export class SigninRequest {
 }
 
 // @public (undocumented)
-export interface SigninRequestArgs {
+export interface SigninRequestCreateArgs {
     // (undocumented)
     acr_values?: string;
     // (undocumented)
@@ -731,22 +732,6 @@ export type SigninSilentArgs = IFrameWindowParams & ExtraSigninRequestArgs;
 
 // @public (undocumented)
 export class SigninState extends State {
-    constructor(args: {
-        id?: string;
-        data?: unknown;
-        created?: number;
-        request_type?: string;
-        url_state?: string;
-        code_verifier?: string | boolean;
-        authority: string;
-        client_id: string;
-        redirect_uri: string;
-        scope: string;
-        client_secret?: string;
-        extraTokenParams?: Record<string, unknown>;
-        response_mode?: "query" | "fragment";
-        skipUserInfo?: boolean;
-    });
     // (undocumented)
     readonly authority: string;
     // (undocumented)
@@ -756,9 +741,11 @@ export class SigninState extends State {
     readonly code_challenge: string | undefined;
     readonly code_verifier: string | undefined;
     // (undocumented)
+    static create(args: SigninStateCreateArgs): Promise<SigninState>;
+    // (undocumented)
     readonly extraTokenParams: Record<string, unknown> | undefined;
     // (undocumented)
-    static fromStorageString(storageString: string): SigninState;
+    static fromStorageString(storageString: string): Promise<SigninState>;
     // (undocumented)
     readonly redirect_uri: string;
     // (undocumented)
@@ -770,6 +757,45 @@ export class SigninState extends State {
     // (undocumented)
     toStorageString(): string;
 }
+
+// @public (undocumented)
+export interface SigninStateArgs {
+    // (undocumented)
+    authority: string;
+    // (undocumented)
+    client_id: string;
+    // (undocumented)
+    client_secret?: string;
+    // (undocumented)
+    code_challenge?: string;
+    // (undocumented)
+    code_verifier?: string;
+    // (undocumented)
+    created?: number;
+    // (undocumented)
+    data?: unknown;
+    // (undocumented)
+    extraTokenParams?: Record<string, unknown>;
+    // (undocumented)
+    id?: string;
+    // (undocumented)
+    redirect_uri: string;
+    // (undocumented)
+    request_type?: string;
+    // (undocumented)
+    response_mode?: "query" | "fragment";
+    // (undocumented)
+    scope: string;
+    // (undocumented)
+    skipUserInfo?: boolean;
+    // (undocumented)
+    url_state?: string;
+}
+
+// @public (undocumented)
+export type SigninStateCreateArgs = Omit<SigninStateArgs, "code_verifier"> & {
+    code_verifier?: string | boolean;
+};
 
 // @public (undocumented)
 export type SignoutPopupArgs = PopupWindowParams & ExtraSignoutRequestArgs;
@@ -838,7 +864,7 @@ export class State {
     readonly created: number;
     readonly data?: unknown;
     // (undocumented)
-    static fromStorageString(storageString: string): State;
+    static fromStorageString(storageString: string): Promise<State>;
     // (undocumented)
     readonly id: string;
     // (undocumented)
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,14 +9,13 @@
       "version": "2.4.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "crypto-js": "^4.2.0",
         "jwt-decode": "^4.0.0"
       },
       "devDependencies": {
         "@microsoft/api-extractor": "^7.35.0",
         "@testing-library/jest-dom": "^6.0.0",
-        "@types/crypto-js": "^4.1.3",
         "@types/jest": "^29.2.3",
+        "@types/node": "^20.8.2",
         "@typescript-eslint/eslint-plugin": "^6.4.1",
         "@typescript-eslint/parser": "^6.4.1",
         "esbuild": "^0.19.5",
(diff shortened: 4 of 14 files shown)


pamapa:
Thanks for the proof of concept, does not look to invasive :-)

The main thing that questions me is:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle -> Secure context required
I mean for testing and production system this is good, but how do we deal with development systems, where we have HTTP and not HTTPS?

43081j:
@pamapa you're in luck i think. from MDN:

> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.

43081j:
could we possibly come up with a path forward if you get time? i could either use this PR or make a new one with whatever approach we settle on. then maybe we just publish it as a pre-release for now?

im happy to then test it for a while in one of our larger projects

pamapa:
the changes i see look pretty good, i need sometime to test it in one of our projects. Maybe you can do the same. I just copy that source into a subdir of the project and direct the imports accordingly...

43081j:
sure ill give it a go

the tests will fail in CI fyi until we switch test runner (jsdom is missing textencoder and what not, its pretty far behind)

pamapa:
works like a charm here, including silent renew...

src/SigninRequest.ts
```diff
@@ -94,6 +93,21 @@ export class SigninRequest {
             client_secret, scope, extraTokenParams,
             skipUserInfo,
         });
+    }
+
+    public async getUrl(): Promise<string> {
+        /* eslint-disable @typescript-eslint/no-unused-vars */
```
pamapa, line 99: properly setup local variables, such that we do not need eslint-disable here

src/SigninRequest.ts
```diff
@@ -105,8 +119,11 @@ export class SigninRequest {
         }
 
         parsedUrl.searchParams.append("state", this.state.id);
-        if (this.state.code_challenge) {
-            parsedUrl.searchParams.append("code_challenge", this.state.code_challenge);
+
+        const challenge = await this.state.getChallenge();
+
```
pamapa, line 124: no need for newline here

src/SigninState.ts
```diff
@@ -72,6 +68,20 @@ export class SigninState extends State {
         this.skipUserInfo = args.skipUserInfo;
     }
 
+    public async getChallenge(): Promise<string|undefined> {
+        if (!this.code_verifier) {
+            return undefined;
+        }
+
+        let challenge = this._code_challenge;
+
```
pamapa, line 77: no need for newline here

src/SigninRequest.ts
43081j: its a slight pain thanks to the fact we append the `optionalParams` to the query string.

if we want to get rid of the unused vars, they'll become part of `optionalParams`.

which means we'd have to instead have a fixed set of properties we append to the qs and hope the value actually matches the type at runtime. it'd be a change in behaviour since random nonsense people pass in at runtime right now will end up in the qs.

src/SigninRequest.ts
```diff
@@ -94,6 +95,38 @@ export class SigninRequest {
             client_secret, scope, extraTokenParams,
             skipUserInfo,
         });
+    }
+
+    public async getUrl(): Promise<string> {
+        const nonOptionalParams: Array<keyof SigninRequestArgs> = [
```
43081j, line 101: this right here is what i meant.

we can't rely on object rest to get `...optionalParams` because it'd contain properties previously not included in that set (the unused vars). 

alternatively, we could flip this around and have an array of optional prop names.

whichever way we do it, we need an array of keys i think. 

the way i have it now, if we pass unknown properties (type errors) at runtime as a js consumer, they'll make it through to the query string.

but i admit whatever we do its not great looking

pamapa: You are right both way are not really nice. This does it more explicit. Also more code... I like this version, but can live with the first version you had with the eslint ignore unused variables...
```js
export class SigninRequest {
    private readonly _logger = new Logger("SigninRequest");

    public readonly state: SigninState;

    private readonly _url: string;
    private readonly _client_id: string;
    private readonly _redirect_uri: string;
    private readonly _response_type: string;
    private readonly _scope: string;
    private readonly _response_mode?: "query" | "fragment";
    private readonly _nonce?: string;
    private readonly _resource?: string | string[];
    private readonly _extraParams: Partial<SigninRequestArgs>;

    public constructor({
        // mandatory
        url, authority, client_id, redirect_uri, response_type, scope,
        // optional
        state_data, response_mode, request_type, client_secret, nonce,
        resource,
        skipUserInfo,
        extraQueryParams,
        extraTokenParams,
        disablePKCE,
        ...extraParams
    }: SigninRequestArgs) {
        if (!url) {
            this._logger.error("ctor: No url passed");
            throw new Error("url");
        }
        if (!client_id) {
            this._logger.error("ctor: No client_id passed");
            throw new Error("client_id");
        }
        if (!redirect_uri) {
            this._logger.error("ctor: No redirect_uri passed");
            throw new Error("redirect_uri");
        }
        if (!response_type) {
            this._logger.error("ctor: No response_type passed");
            throw new Error("response_type");
        }
        if (!scope) {
            this._logger.error("ctor: No scope passed");
            throw new Error("scope");
        }
        if (!authority) {
            this._logger.error("ctor: No authority passed");
            throw new Error("authority");
        }

        this._url = url;
        this._client_id = client_id;
        this._redirect_uri = redirect_uri;
        this._response_type = response_type;
        this._scope = scope;
        this._response_mode = response_mode;
        this._client_id = client_id;
        this._nonce = nonce;
        this._resource = resource;
        this._extraParams = extraParams;

        this.state = new SigninState({
            data: state_data,
            request_type,
            code_verifier: !disablePKCE,
            client_id, authority, redirect_uri,
            response_mode,
            client_secret, scope, extraTokenParams,
            skipUserInfo,
        });
    }

    public async getUrl(): Promise<string> {
        const parsedUrl = new URL(this._url);
        parsedUrl.searchParams.append("client_id", this._client_id);
        parsedUrl.searchParams.append("redirect_uri", this._redirect_uri);
        parsedUrl.searchParams.append("response_type", this._response_type);
        parsedUrl.searchParams.append("scope", this._scope);
        if (this._nonce) {
            parsedUrl.searchParams.append("nonce", this._nonce);
        }

        parsedUrl.searchParams.append("state", this.state.id);

        const challenge = await this.state.getChallenge();
        if (challenge) {
            parsedUrl.searchParams.append("code_challenge", challenge);
            parsedUrl.searchParams.append("code_challenge_method", "S256");
        }

        if (this._resource) {
            // https://datatracker.ietf.org/doc/html/rfc8707
            const resources = Array.isArray(this._resource) ? this._resource : [this._resource];
            resources
                .forEach(r => parsedUrl.searchParams.append("resource", r));
        }

        for (const [key, value] of Object.entries({ response_mode: this._response_mode, ...this._extraParams })) {
            if (value != null) {
                parsedUrl.searchParams.append(key, value.toString());
            }
        }

        return parsedUrl.href;
    }
}
```

43081j: thats true, just putting them on `this` individually. i think i'd prefer that over relying on unused vars as it seems they could fall over/be a bit flaky over time.

ill give it a try

pamapa:
My plan is to have a 3.0.0 in Q3/Q4 of this year. The other thing i would like to land is the fix of the current strange claims merge behavior...

43081j:
we still need to move over to web-test-runner too, which is why this PR is blocked

its a fairly big move i think

pamapa:
> we still need to move over to web-test-runner too, which is why this PR is blocked

Do you have time to adapt the unit-tests as well sometime in the next weeks?

43081j:
i started doing it on a branch but it kinda spiralled. ill probably throw that away and try again from scratch

will make sure i have a go at it soon so we can at least know how big the task is

43081j:
@pamapa just to update you on this

the big chunk of work to migrate to WTR seems to actually be migration off jest's assertions and mocks.

migrating jest assertions to chai is straight forward but time consuming (lot of find & replace).

we then have to migrate off the mocks to hanbi (used by modernweb) or sinon

alternatively we try figure out a way to use jest for its assertions/mocks but not its test runner. feels sketchy going half way like that, though...

kitten:
Hiya 👋 
I don't want to be too pushy here or hurry this along, but just offer a commit for the missing `TextEncoder` test setup and mocks for Jest: https://github.com/authts/oidc-client-ts/commit/4df31ec34ce9808121d85ed864fb3cb0e56239e3

Totally fine to ignore this, but I thought I'd just post and mention it here, in case it was blocking and this could help this draft PR along :v:

kitten:
Are you referring to writing "real world" browser E2E/integration tests by any chance?

I've been looking into the same a while back for another OSS project, and the options aren't great just yet. The most sensible has been Cypress + Component testing + Vite imho. If that's indeed though what you're looking for, I wish [Vitest's Browser Mode](https://vitest.dev/guide/browser.html) was ready already, since it'd be perfect for mixing browser- and unit-tests.

Just to add an opinion, not sure if that's the task at hand for this repo, but if it's a big blocker, I'd suggest not throwing away all unit tests and instead using any browser test as a supplement, to not start from scratch entirely 🤔

Anyway, I'll get out of your hair now. No rush & Cheers! ❤️

43081j:
> Are you referring to writing "real world" browser E2E/integration tests by any chance?
> 
> I've been looking into the same a while back for another OSS project, and the options aren't great just yet. The most sensible has been Cypress + Component testing + Vite imho. If that's indeed though what you're looking for, I wish [Vitest's Browser Mode](https://vitest.dev/guide/browser.html) was ready already, since it'd be perfect for mixing browser- and unit-tests.
> 
> Just to add an opinion, not sure if that's the task at hand for this repo, but if it's a big blocker, I'd suggest not throwing away all unit tests and instead using any browser test as a supplement, to not start from scratch entirely 🤔
> 
> Anyway, I'll get out of your hair now. No rush & Cheers! ❤️

[web-test-runner](https://github.com/modernweb-dev/web/tree/master/packages/test-runner) is basically one of the go-to solutions these days around the web-first/native community.

its very well tested, widely used, etc. super easy to get setup

the problem is more that the current tests here were built in a node-only environment originally. which is the reason jest works (jest is node-only itself).

there's various things you can do like transforming CJS on test run (to ESM), and polyfilling node APIs (path/fs/etc). but its all really avoiding the reality that is jest isn't made to work in a browser, we should be using a more modern test _assertion_ and _mocking_ library (WTR handles the test runner part already).

there's already a few options for that. the obvious ones are chai (we recently published an ESM-only version of chai that works natively in browsers) and sinon or hanbi (both should work in browsers, sinon has a much larger footprint).

fwiw too vitest will one day be in a similar place to wtr but ultimately is a thin layer around vite, which itself is a thin layer around rollup. so we'd run into the exact same issues as here

pamapa:
this simple patch makes this merge request work:
```patch
 test/setup.ts | 15 +++++++++++++++
 1 file changed, 15 insertions(+)

diff --git a/test/setup.ts b/test/setup.ts
index 34b5b46..8178605 100644
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -1,6 +1,21 @@
+import { TextEncoder } from "util";
+
 import { Log } from "../src";
 
 beforeAll(() => {
+    // mock features which are not present globally in jsdom
+    globalThis.TextEncoder = TextEncoder;
+    Object.defineProperty(globalThis, "crypto", {
+        value: {
+            getRandomValues: jest.fn(),
+            subtle: {
+                digest: jest.fn().mockImplementation(
+                    () => [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+                ),
+            },
+        },
+    });
+
     globalThis.fetch = jest.fn();
 
     const unload = () =>
```

The main issue is that `testEnvironment: "jsdom"` does not yet support `subtle` and `getRandomValues`. For unit-tests proposes it is the easiest to simply mock them...

43081j:
> this simple patch makes this merge request work

we were already aware of this long ago, that isn't the blocker here.

the whole point of this move to built-in crypto was to make oidc-client-ts work in a browser natively. this PR already does that near enough. the last piece of this PR is to then test the module in a real browser.

the blocker here is that. we want to test the code in a real browser, i.e. use web-test-runner to run our unit tests. 

inevitably it'll have to happen, in this PR or a followup PR. one way or another some non-jest tests need setting up, against a real browser (if you want any confidence in what is being shipped).

43081j:
to be clear:

1. I don't think anyone was suggesting this. I'm aware it is black box and you are too, neither of us were wanting to test it any other way
2. it would be the exact same unit tests as you have now but in a real browser environment rather than a (outdated) mock one. integration/e2e tests would be a different suite of tests. right now im only talking about running the exact same _unit_ tests (definitely unit tests) in node and the browser (both so we know the library passes tests in both envs)

happy to cut the corner in this PR if thats what you want, though

and a follow up PR at some point can introduce WTR for running the unit tests, probably alongside jest so we still test against a node env

pamapa:
Thanks for the clarification, I agree with you.

Can you please open up a new issue about modernizing the existing unit-test bases on jest. Lets have the discussion there.

This merge request shall persevere the current unit-test framework as is...

43081j:
Sure that makes sense 👍

Will try find time this weekend

43081j:
- imported textencoder from node and put it on globalThis
- imported webcrypto from node and assigned its subtle to `globalThis.crypto` (since jsdom puts its own implementation there which is missing subtle)

let me know what you think

not entirely sure of the node compat either, i.e. what the oldest supported version would be

pamapa:
> not entirely sure of the node compat either, i.e. what the oldest supported version would be

https://nodejs.org/docs/latest-v15.x/api/webcrypto.html#webcrypto_crypto_subtle
-> first LTS after 15 would be 16

43081j:
@pamapa any way we can move this forward? :eyes:

pamapa:
I planning to do a 2.2.5 release sometime in the next days/week, when that is gone we can use `main` for 3.0.0 beta releases...

src/nodeCrypto.ts
```diff
@@ -0,0 +1,5 @@
+declare module "node:crypto" {
```
huysentruitw, line 1: File not required I guess?

43081j:
i've also removed the webcrypto types i had, and instead added `@types/node` as a direct dev dependency (which now has the webcrypto types).

it looks like builds worked before because jest pulled `@types/node` in and we picked it up from there. really it should've been a dependency all along

pamapa:
lint error in CI:
```
Run npm run lint

> oidc-client-ts@2.3.0 lint


/home/runner/work/oidc-client-ts/oidc-client-ts/src/SigninRequest.ts
Error:   160:52  error  'value' will evaluate to '[object Object]' when stringified  @typescript-eslint/no-base-to-string

✖ 1 problem (1 error, 0 warnings)

Error: Process completed with exit code 1.
```

pamapa:
> interesting, its because technically `value` could be an object but should be a primitive
> 
> i've added a tiny `isPrimitive` type guard in that file to get the type right

Please not, you can simply declare the type different e.g.:

simple version:
```js
private readonly _optionalParams: Record<string, string | number> | undefined;
```

more code version:
```js
private readonly _optionalParams: Partial<Omit<SigninRequestArgs,"state_data" | "extraQueryParams" | "extraTokenParams">>;
```

And also add a space before and after the pipe:
e.g.
`_resource: string|string[]|undefined` -> `_resource: string | string[] | undefined`, ... there are more like this

And `_authority` seems unused -> lets drop it

43081j:
The omit list is a bit longer but I'll put the change in 👍

Just didn't before in case you didn't want the property types changing

43081j:
there you go @pamapa

pamapa:
alternative, than storing all whats needed for `getUrl` we simple store the almost calculated url, execpt the async `challenge`, which can/must still be done in `getUrl`.

Something like:
```ts
public async getUrl(): Promise<string> {
  const parsedUrl = this._preparedUrl; // contains all other stuff already done in constructor
  
  const challenge = await this.state.getChallenge();
  if (challenge) {
    parsedUrl.searchParams.append("code_challenge", challenge);
    parsedUrl.searchParams.append("code_challenge_method", "S256");
  }

  return parsedUrl.href;
}
```

pamapa:
> Seems like `getUrl` is added because of the challenge that needs to be generated async instead, right?

yes

> Can we use the [existing async-factory method](https://github.com/authts/oidc-client-ts/blob/f3a805aa00a021a8844498df8a3c8dd64c7c2f7b/src/OidcClient.ts#L86) here to generate a SignInRequest instance and move the logic from the SignInRequest constructor to the factory method instead?

sounds great, instead of the `async getUrl`, we can add a `async setUrl`, which calculates the url in `createSigninRequest`, the access to the url can stay as its is...

43081j:
@pamapa what do you think of the last commit?

do you wanna do it a different way? i did remove `url` from the request params since the constructor no longer needed it this way

src/UserManager.ts
```diff
@@ -424,7 +424,11 @@ export class UserManager {
             const signinRequest = await this._client.createSigninRequest(args);
             logger.debug("got signin request");
 
-            const url = await signinRequest.getUrl();
+            const url = signinRequest.url;
+
+            if (!url) {
```
pamapa, line 429: it guess its more safe to move the throw in the `url` getter and return always a `string`. This way we do not need any change in this file...

pamapa:
> @pamapa what do you think of the last commit?
> 
> do you wanna do it a different way? i did remove `url` from the request params since the constructor no longer needed it this way

looks good

src/UserManager.ts
43081j: as far as the type system's concerned, you could forget to call `setUrl`, though. which is why `this.url` can be `undefined`

we happen to know we call it, though. so maybe we could just default `this._url` to `''` for sake of it always being a string (since we'll have always called `setUrl` to replace that anyway)

43081j: ah nevermind, i get you! i misread 👍

huysentruitw:
I have made some changes in a Draft PR https://github.com/authts/oidc-client-ts/pull/1196 because it was too much to add as review comments to this PR. Excuse me about that.

But in my draft, the changes should be less intrusive, as I just created async factory methods in SigninState & SigninRequest, so we can leave the rest as-is. 

@43081j  feel free to cherry pick my latest commit into your PR here.

43081j:
rebased and done the refactor

jimmyvdberg-effectory:
Would be lovely if you could move forward with this. There is now also a vulnerability found in crypto-js about using a weak algorithm.  More info can be found here: [https://security.snyk.io/vuln/SNYK-JS-CRYPTOJS-6028119](https://security.snyk.io/vuln/SNYK-JS-CRYPTOJS-6028119)

pamapa:
> Would be lovely if you could move forward with this. There is now also a vulnerability found in crypto-js about using a weak algorithm. More info can be found here: https://security.snyk.io/vuln/SNYK-JS-CRYPTOJS-6028119

No worries yet, we do **not** use `CryptoJS.PBKDF2`. We use only this:
```
import CryptoJS from "crypto-js/core.js";
import sha256 from "crypto-js/sha256.js";
import Base64 from "crypto-js/enc-base64.js";
import Utf8 from "crypto-js/enc-utf8.js";
```

But yes its now deprecated (thanks for the hint), it makes sense to proceed.

pamapa:
@@43081j thanks for the rebase. Whats holding back merging? I guess you synced with 1196. from @huysentruitw # Can we remove the "wip: " in the title? My plan is to merge this next week...

43081j:
i think all should be fine, it does use async factory methods now as suggested

it'd be good if you could double check the change in public interface (i.e. `docs/` changes). i basically introduced a `SigninStateArgs` instead of having the type inlined in the constructor, since i needed to reuse it at the time

pamapa:
@43081j Small things:

As we now have `create`, lets make the constructor private, such that create must be used and lets align the type name to the new create function (`SigninRequestArgs` -> `SigninRequestCreateArgs`) so it matches...
```
diff --git a/src/OidcClient.test.ts b/src/OidcClient.test.ts
index 577d239..aa3d43d 100644
--- a/src/OidcClient.test.ts
+++ b/src/OidcClient.test.ts
@@ -263,15 +263,15 @@ describe("OidcClient", () => {
 
         it("should deserialize stored state and return state and response", async () => {
             // arrange
-            const item = new SigninState({
+            const item = await SigninState.create({
                 id: "1",
                 authority: "authority",
                 client_id: "client",
                 redirect_uri: "http://app/cb",
                 scope: "scope",
                 request_type: "type",
-            }).toStorageString();
-            jest.spyOn(subject.settings.stateStore, "get").mockImplementation(() => Promise.resolve(item));
+            });
+            jest.spyOn(subject.settings.stateStore, "get").mockImplementation(() => Promise.resolve(item.toStorageString()));
 
             // act
             const { state, response } = await subject.readSigninResponseState("http://app/cb?state=1");
@@ -318,7 +318,7 @@ describe("OidcClient", () => {
 
         it("should deserialize stored state and call validator", async () => {
             // arrange
-            const item = new SigninState({
+            const item = await SigninState.create({
                 id: "1",
                 authority: "authority",
                 client_id: "client",
diff --git a/src/OidcClient.ts b/src/OidcClient.ts
index 652a67d..f776950 100644
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -7,7 +7,7 @@ import { type OidcClientSettings, OidcClientSettingsStore } from "./OidcClientSe
 import { ResponseValidator } from "./ResponseValidator";
 import { MetadataService } from "./MetadataService";
 import type { RefreshState } from "./RefreshState";
-import { SigninRequest, type SigninRequestArgs } from "./SigninRequest";
+import { SigninRequest, type SigninRequestCreateArgs } from "./SigninRequest";
 import { SigninResponse } from "./SigninResponse";
 import { SignoutRequest, type SignoutRequestArgs } from "./SignoutRequest";
 import { SignoutResponse } from "./SignoutResponse";
@@ -20,7 +20,7 @@ import { ClaimsService } from "./ClaimsService";
  * @public
  */
 export interface CreateSigninRequestArgs
-    extends Omit<SigninRequestArgs, "url" | "authority" | "client_id" | "redirect_uri" | "response_type" | "scope" | "state_data"> {
+    extends Omit<SigninRequestCreateArgs, "url" | "authority" | "client_id" | "redirect_uri" | "response_type" | "scope" | "state_data"> {
     redirect_uri?: string;
     response_type?: string;
     scope?: string;
diff --git a/src/SigninRequest.test.ts b/src/SigninRequest.test.ts
index 8708e48..a2e779d 100644
--- a/src/SigninRequest.test.ts
+++ b/src/SigninRequest.test.ts
@@ -1,13 +1,13 @@
 // Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-import { SigninRequest, type SigninRequestArgs } from "./SigninRequest";
+import { SigninRequest, type SigninRequestCreateArgs } from "./SigninRequest";
 import { URL_STATE_DELIMITER } from "./utils";
 
 describe("SigninRequest", () => {
 
     let subject: SigninRequest;
-    let settings: SigninRequestArgs;
+    let settings: SigninRequestCreateArgs;
 
     beforeEach(async () => {
         settings = {
diff --git a/src/SigninRequest.ts b/src/SigninRequest.ts
index 0b8ebfb..b44bfb5 100644
--- a/src/SigninRequest.ts
+++ b/src/SigninRequest.ts
@@ -8,7 +8,7 @@ import { SigninState } from "./SigninState";
  * @public
  * @see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
  */
-export interface SigninRequestArgs {
+export interface SigninRequestCreateArgs {
     // mandatory
     url: string;
     authority: string;
@@ -54,7 +54,7 @@ export class SigninRequest {
     public readonly url: string;
     public readonly state: SigninState;
 
-    public constructor(args: {
+    private constructor(args: {
         url: string;
         state: SigninState;
     }) {
@@ -73,7 +73,7 @@ export class SigninRequest {
         extraTokenParams,
         disablePKCE,
         ...optionalParams
-    }: SigninRequestArgs): Promise<SigninRequest> {
+    }: SigninRequestCreateArgs): Promise<SigninRequest> {
         if (!url) {
             this._logger.error("create: No url passed");
             throw new Error("url");
diff --git a/src/SigninState.ts b/src/SigninState.ts
index 57c0d80..a53be0e 100644
--- a/src/SigninState.ts
+++ b/src/SigninState.ts
@@ -57,7 +57,7 @@ export class SigninState extends State {
 
     public readonly skipUserInfo: boolean | undefined;
 
-    public constructor(args: SigninStateArgs) {
+    private constructor(args: SigninStateArgs) {
         super(args);
 
         this.code_verifier = args.code_verifier;
diff --git a/src/index.ts b/src/index.ts
index 322bcaa..387328c 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,7 @@ export type { OidcClientSettings, SigningKey, ExtraHeader } from "./OidcClientSe
 export type { OidcMetadata } from "./OidcMetadata";
 export { SessionMonitor } from "./SessionMonitor";
 export type { SessionStatus } from "./SessionStatus";
-export type { SigninRequest, SigninRequestArgs } from "./SigninRequest";
+export type { SigninRequest, SigninRequestCreateArgs } from "./SigninRequest";
 export type { RefreshState } from "./RefreshState";
 export { SigninResponse } from "./SigninResponse";
 export { SigninState } from "./SigninState";
```

Would be nice if you could apply this patch on-top/into to your commit...

pamapa:
@43081j and @huysentruitw awesome thanks for contributing and taking care!

pamapa (approved):

PR #917, opened 2023-03-09
Add static or dynamic custom header for each client calls.

BROUSSOLLE-Brice:
New feature

I've made this PR because I need to have custom headers for the client calls.
Our DevOps team need those custom headers to have better reading and filtering logs on low level network components.

I'm not sure if I've done code as your standards. I'm fully opened to share and discuss about it. 
Hope this feature could help other people and accepted by you.

### Changes

New optional setting `customHeaders` into `OidcClientSettings`.

`customHeaders` is an hash array with header name as a key. Values could be string or a function returning a string.

Some headers are protected:
 - Authorization
 - Accept
 - Content-Type

### Checklist

### DIFF
--- a/src/JsonService.ts
+++ b/src/JsonService.ts
@@ -10,6 +10,11 @@ import { Logger } from "./utils";
 export type JwtHandler = (text: string) => Promise<Record<string, unknown>>;
 
 /**
+ * @public
+ */
+export type CustomHeader = string | (() => string);
+
+/**
  * @internal
  */
 export interface GetJsonOpts {
@@ -30,6 +35,33 @@ export interface PostFormOpts {
 /**
  * @internal
  */
+function appendCustomHeaders(
+    headers: Record<string, string>,
+    customs: Record<string, CustomHeader>,
+): void {
+    const customKeys = Object.keys(customs);
+    const protectedHeaders = [
+        "authorization",
+        "accept",
+        "content-type",
+    ];
+    if (customKeys.length === 0) {
+        return;
+    }
+    customKeys.forEach((headerName) => {
+        if (protectedHeaders.includes(headerName.toLocaleLowerCase())) {
+            return;
+        }
+        const content = (typeof customs[headerName] === "function") ?
+            (customs[headerName] as ()=>string)() :
+            customs[headerName];
+        headers[headerName] = content as string;
+    });
+}
+
+/**
+ * @internal
+ */
 export class JsonService {
     private readonly _logger = new Logger("JsonService");
 
@@ -38,6 +70,7 @@ export class JsonService {
     public constructor(
         additionalContentTypes: string[] = [],
         private _jwtHandler: JwtHandler | null = null,
+        private _customHeaders: Record<string, CustomHeader> = {},
     ) {
         this._contentTypes.push(...additionalContentTypes, "application/json");
         if (_jwtHandler) {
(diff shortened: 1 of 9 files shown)


src/JsonService.ts
```diff
@@ -27,6 +32,35 @@ export interface PostFormOpts {
     initCredentials?: "same-origin" | "include" | "omit";
 }
 
+/**
+ * @internal
+ */
+function appendCustomHeaders(
```
pamapa, line 38: please make this an private function within JsonService class

src/OidcClientSettings.ts
```diff
@@ -130,6 +131,11 @@ export interface OidcClientSettings {
      * Only scopes in this list will be passed in the token refresh request.
      */
     refreshTokenAllowedScope?: string | undefined;
+
+    /**
+     * Set additional custom headers to be passed to the client
+     */
+    customHeaders?: Record<string, CustomHeader>;
```
pamapa, line 138: lets name this `extraHeaders` as we already have other extra stuff like: `extraQueryParams` and `extraTokenParams`

src/JsonService.ts
BROUSSOLLE-Brice: I've made the modification in my last commit.

src/JsonService.ts
```diff
@@ -136,6 +144,7 @@ export class JsonService {
         if (basicAuth !== undefined) {
             headers["Authorization"] = "Basic " + basicAuth;
         }
+        this.appendExtraHeaders(headers);
```
pamapa, line 147: add a space between the if block and this new code

src/JsonService.ts
```diff
@@ -177,4 +186,29 @@ export class JsonService {
...
+            "authorization",
+            "accept",
+            "content-type",
+        ];
+        if (customKeys.length === 0) {
+            return;
+        }
+        customKeys.forEach((headerName) => {
+            if (protectedHeaders.includes(headerName.toLocaleLowerCase())) {
+                return;
```
pamapa, line 204: at least log an warning here

src/OidcClientSettings.ts
```diff
@@ -209,6 +216,7 @@ export class OidcClientSettingsStore {
         // extra query params
```
pamapa, line 216: `// extra query params` -> `// extra`

src/JsonService.ts
```diff
@@ -9,6 +9,11 @@ import { Logger } from "./utils";
  */
 export type JwtHandler = (text: string) => Promise<Record<string, unknown>>;
 
+/**
+ * @public
+ */
+export type ExtraHeader = string | (() => string);
```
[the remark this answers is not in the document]
BROUSSOLLE-Brice: Hi, 
1. Yes for me, I need to have a function () => string for dynamic values during the app run. Not only on init of the app.
2. Ok

src/JsonService.ts
BROUSSOLLE-Brice: Ok

src/JsonService.ts
BROUSSOLLE-Brice: ok

src/JsonService.ts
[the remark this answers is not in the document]
pamapa: 1. we can do it like its done for the other settings (e.g. `extraQueryParams`), we add them to the `ExtraSigninRequestArgs`, then you can pass it dynamically via `signinRedirect`, ...

src/JsonService.ts
BROUSSOLLE-Brice: `logger.warn("Protected header could not be overridden", headerName, protectedHeaders);`
 
It's good for you ?

pamapa (approved):

PR #977, opened 2023-04-21
fix: included max_age in sign in args

### LINKED ISSUE #976: Ability to override max_age for each signIn request
Currently `max_age` can only be specified as an option to `UserManager`, but we need to be able to specify it per `signIn` call

The reason for this is for implementing an [OIDC re-authentication flow](https://auth0.com/docs/authenticate/login/max-age-reauthentication) for specific feature of the application.

Desired behaviour:
1. User login to the application via OIDC provider - no `max_age` parameter should be sent as we want to have SSO
2. User uses the specific high risk feature, and we want to make sure that it is still the same user at the keyboard. So the app would do a `signinRedirect` or `signinPopup`, with `max_age` set to a relative small period (e.g. 5 minutes).

jackie-linz:
Closes/fixes #976

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -91,7 +91,7 @@ export class ErrorTimeout extends Error {
 export type ExtraHeader = string | (() => string);
 
 // @public (undocumented)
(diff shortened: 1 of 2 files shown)


philals (approved):

pamapa (approved):

pamapa:
thanks for the contribution!

PR #982, opened 2023-04-24
Allow auth_time to be undefined on token refresh

### LINKED ISSUE #772: auth_time does not exist in refresh id_token (PingID / Azure AD)
I have a vanilla javascript application that uses PingID to authenticate with an Azure AD tenant.

I get the following error on token expiry or when I manually invoke `mgr.signinSilent()`:

```
auth_time in id_token does not match original auth_time
```

After comparing the original `id_token` and the refreshed one I can see that `auth_time` is missing in the refreshed token. Other properties of the original token lead me to believe I'm going through some kind of SAML <-> OIDC marshalling for my initial token and a slightly different method on refresh.

Could this validation check only operate if the field is present? For example:

```javascript
if (current.auth_time && current.auth_time !== profile.auth_time) {
    logger.throw(new Error("auth_time in id_token does not match original auth_time"));
}
```

to

```javascript
if (current.auth_time && profile.auth_time !== undefined && current.auth_time !== profile.auth_time) {
    logger.throw(new Error("auth_time in id_token does not match original auth_time"));
}
```

cestorer:
Closes/fixes #772

Per [response spec](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse), auth_time is conditional dependent on max_age being provided in the request. To fully implement the spec regarding auth_time we would also need access to the request max_age being provided to understand if auth_time is required.

Also renamed variables to assist with logic evaluation. It's likely parameters were being misused based on error messaging, although still valid.

### Checklist

### DIFF
--- a/src/ResponseValidator.ts
+++ b/src/ResponseValidator.ts
@@ -186,28 +186,28 @@ export class ResponseValidator {
         }
     }
 
-    protected _validateIdTokenAttributes(response: SigninResponse, currentToken?: string): void {
+    protected _validateIdTokenAttributes(response: SigninResponse, existingToken?: string): void {
         const logger = this._logger.create("_validateIdTokenAttributes");
 
         logger.debug("decoding ID Token JWT");
-        const profile = JwtUtils.decode(response.id_token ?? "");
+        const incoming = JwtUtils.decode(response.id_token ?? "");
 
-        if (!profile.sub) {
+        if (!incoming.sub) {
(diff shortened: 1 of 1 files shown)


pamapa:
nice work, does not compile yet: `Error: src/ResponseValidator.ts(215,28): error TS2304: Cannot find name 'profile'.`

cestorer:
Dumb mistake - incoming fix commit

pamapa:
thanks for contributing

PR #1058, opened 2023-06-21
fix: remove state attribute from sample requests

demkom58:
If I understand right, here is no reason to specify state, since it will be automatically generated and specified value will be overwritten, it is really misleading.

### DIFF
--- a/samples/Parcel/src/oidc-client/sample.js
+++ b/samples/Parcel/src/oidc-client/sample.js
@@ -36,7 +36,10 @@ var client = new OidcClient(settings);
 // functions for UI elements
 ///////////////////////////////
 function signin() {
-    client.createSigninRequest({ state: { bar: 15 } }).then(function(req) {
+    var optionalArgs = { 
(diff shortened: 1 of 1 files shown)


pamapa:
you are right those states make no sense like they are. They are there to demonstrate the feature though. Maybe its better to have something like:
```js
var optionalArgs = { state: { bar: 15 } }; // demonstrating how to use custom state
client.createSigninRequest(args).then(function(req) {
```

demkom58:
> you are right those states make no sense like they are. They are there to demonstrate the feature though. Maybe its better to have something like:
> 
> ```js
> var optionalArgs = { state: { bar: 15 } }; // demonstrating how to use custom state
> client.createSigninRequest(args).then(function(req) {
> ```

Oh, I thought that this is a state that should be checked by the client for correspondence with the sent one, but it seems that it is rather the data that is sent in the parameters, haha. 🥲 

Might be worth pointing it out in the comments.

pamapa (approved):

pamapa:
thanks for contributing

PR #1148, opened 2023-08-28
Feature: overriding Navigators for non Window browser envs

### LINKED ISSUE #1147: Adding support for custom navigators?
Would the maintainers be open to another PR focused at adding support for custom navigators? (Kinda did ok on the storage engine one, working well in a product with Electron!)

Scenario:

Building a React + Ionic + Capacitor app for iOS. It kinda works ok if you let it leave the app and redirect back.

Ideal solution is to use the Capacitor Browser plug-in to open it without leaving the app.

My idea:

Add 3 new settings to "UserManagerSettings". 

That either get defaulted to the existing if not defined like here:

https://github.com/authts/oidc-client-ts/blob/main/src/UserManagerSettings.ts#L187

Then alter here to use the values from "UserManagerSettings":

https://github.com/authts/oidc-client-ts/blob/main/src/UserManager.ts#L98

philjones88:
Fixes #1147 

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -131,7 +131,7 @@ export interface ILogger {
 // @public (undocumented)
 export interface INavigator {
     // (undocumented)
-    callback(url: string, keepOpen?: boolean): Promise<void>;
+    callback(url: string, params?: unknown): Promise<void>;
     // Warning: (ae-forgotten-export) The symbol "IWindow" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -952,7 +952,7 @@ export type UserLoadedCallback = (user: User) => Promise<void> | void;
 
 // @public
 export class UserManager {
-    constructor(settings: UserManagerSettings);
+    constructor(settings: UserManagerSettings, redirectNavigator?: RedirectNavigator, popupNavigator?: PopupNavigator, iframeNavigator?: IFrameNavigator);
     // (undocumented)
     protected _buildUser(signinResponse: SigninResponse, verifySub?: string): Promise<User>;
     clearStaleState(): Promise<void>;
@@ -963,11 +963,17 @@ export class UserManager {
     protected readonly _events: UserManagerEvents;
     getUser(): Promise<User | null>;
     // (undocumented)
+    protected readonly _iframeNavigator: IFrameNavigator;
+    // (undocumented)
     protected _loadUser(): Promise<User | null>;
     // (undocumented)
     protected readonly _logger: Logger;
     get metadataService(): MetadataService;
+    // (undocumented)
+    protected readonly _popupNavigator: PopupNavigator;
     querySessionStatus(args?: QuerySessionStatusArgs): Promise<SessionStatus | null>;
+    // (undocumented)
+    protected readonly _redirectNavigator: RedirectNavigator;
     removeUser(): Promise<void>;
     // (undocumented)
     protected _revokeInternal(user: User | null, types?: ("access_token" | "refresh_token")[]): Promise<void>;
@@ -1055,7 +1061,6 @@ export interface UserManagerSettings extends OidcClientSettings {
     accessTokenExpiringNotificationTimeInSeconds?: number;
     automaticSilentRenew?: boolean;
     checkSessionIntervalInSeconds?: number;
-    iframeNavigator?: IFrameNavigator;
     iframeNotifyParentOrigin?: string;
     iframeScriptOrigin?: string;
     includeIdTokenInSilentRenew?: boolean;
@@ -1066,13 +1071,11 @@ export interface UserManagerSettings extends OidcClientSettings {
     // (undocumented)
     popup_post_logout_redirect_uri?: string;
     popup_redirect_uri?: string;
-    popupNavigator?: PopupNavigator;
     popupWindowFeatures?: PopupWindowFeatures;
     popupWindowTarget?: string;
     // (undocumented)
     query_status_response_type?: string;
     redirectMethod?: "replace" | "assign";
-    redirectNavigator?: RedirectNavigator;
     redirectTarget?: "top" | "self";
     revokeTokensOnSignout?: boolean;
     revokeTokenTypes?: ("access_token" | "refresh_token")[];
@@ -1094,8 +1097,6 @@ export class UserManagerSettingsStore extends OidcClientSettingsStore {
     // (undocumented)
     readonly checkSessionIntervalInSeconds: number;
     // (undocumented)
-    readonly iframeNavigator: IFrameNavigator;
-    // (undocumented)
     readonly iframeNotifyParentOrigin: string | undefined;
     // (undocumented)
     readonly iframeScriptOrigin: string | undefined;
@@ -1112,8 +1113,6 @@ export class UserManagerSettingsStore extends OidcClientSettingsStore {
     // (undocumented)
     readonly popup_redirect_uri: string;
     // (undocumented)
-    readonly popupNavigator: PopupNavigator;
-    // (undocumented)
     readonly popupWindowFeatures: PopupWindowFeatures;
     // (undocumented)
     readonly popupWindowTarget: string;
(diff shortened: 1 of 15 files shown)


src/UserManager.ts
```diff
@@ -578,7 +570,7 @@ export class UserManager {
     /**
      * Returns promise to trigger a silent request (via an iframe) to the end session endpoint.
      */
-    public async signoutSilent(args: SignoutSilentArgs = {}): Promise<void> {
+    public async signoutSilent(args: SignoutSilentArgs = { silentRequestTimeoutInSeconds: this.settings.silentRequestTimeoutInSeconds }): Promise<void> {
```
pamapa, line 573: That should not needed here `silentRequestTimeoutInSeconds: this.settings.silentRequestTimeoutInSeconds`,
its already done here: `IFrameNavigator`:
```
public async prepare({
        silentRequestTimeoutInSeconds = this._settings.silentRequestTimeoutInSeconds,
```

pamapa:
looks otherwise really nice

src/UserManager.ts
philjones88: Yup, sorry I didn't revert that change by mistake. done.

src/UserManager.ts
```diff
@@ -578,7 +570,7 @@ export class UserManager {
     /**
      * Returns promise to trigger a silent request (via an iframe) to the end session endpoint.
      */
-    public async signoutSilent(args: SignoutSilentArgs = {}): Promise<void> {
+    public async signoutSilent(args: SignoutSilentArgs = { }): Promise<void> {
```
pamapa, line 573: there is still a space left, funny that eslint did not see that :-), would be nice if you can remove that as well

philjones88: fixed

pamapa:
Sorry for the late comment, but i think its better when we do it in a similar way we already have done in OidcClient:
```js
    public constructor(settings: OidcClientSettings);
    public constructor(settings: OidcClientSettingsStore, metadataService: MetadataService); 
    public constructor(settings: OidcClientSettings | OidcClientSettingsStore, metadataService?: MetadataService) {
        this.settings = settings instanceof OidcClientSettingsStore ? settings : new OidcClientSettingsStore(settings);

        this.metadataService = metadataService ?? new MetadataService(this.settings);
```
-> something like
```js
public constructor(settings: UserManagerSettings);
    public constructor(settings: OidcClientSettings | OidcClientSettingsStore, redirectNavigator?: RedirectNavigator, ...) {
        this.settings = settings instanceof UserManagerSettingsStore? settings : new UserManagerSettingsStore(settings);
        ...
        this._redirectNavigator= redirectNavigator?? new RedirectNavigator(this.settings);
```
That allows you to override the `UserManager` class in your code and inject your navigators too without going to extend the settings for this special cases. And its more like an SDK then...

philjones88:
> Sorry for the late comment, but i think its better when we do it in a similar way we already have done in OidcClient:
> 
> ```js
>     public constructor(settings: OidcClientSettings);
>     public constructor(settings: OidcClientSettingsStore, metadataService: MetadataService); 
>     public constructor(settings: OidcClientSettings | OidcClientSettingsStore, metadataService?: MetadataService) {
>         this.settings = settings instanceof OidcClientSettingsStore ? settings : new OidcClientSettingsStore(settings);
> 
>         this.metadataService = metadataService ?? new MetadataService(this.settings);
> ```
> 
> -> something like
> 
> ```js
> public constructor(settings: UserManagerSettings);
>     public constructor(settings: OidcClientSettings | OidcClientSettingsStore, redirectNavigator?: RedirectNavigator, ...) {
>         this.settings = settings instanceof UserManagerSettingsStore? settings : new UserManagerSettingsStore(settings);
>         ...
>         this._redirectNavigator= redirectNavigator?? new RedirectNavigator(this.settings);
> ```
> 
> That allows you to override the `UserManager` class in your code and inject your navigators too without going to extend the settings for this special cases. And its more like an SDK then...

Would you suggest:

```
    public constructor(settings: UserManagerSettings);
    public constructor(settings: UserManagerSettings, redirectNavigator?: RedirectNavigator);
    public constructor(settings: UserManagerSettings, redirectNavigator?: RedirectNavigator, popupNavigator?: PopupNavigator);
    public constructor(settings: UserManagerSettings, redirectNavigator?: RedirectNavigator, popupNavigator?: PopupNavigator, iframeNavigator?: IFrameNavigator) {
```

or

```
    public constructor(settings: UserManagerSettings);
    public constructor(settings: UserManagerSettings, redirectNavigator?: RedirectNavigator, popupNavigator?: PopupNavigator, iframeNavigator?: IFrameNavigator) {
```

philjones88:
> Sorry for the late comment, but i think its better when we do it in a similar way we already have done in OidcClient:
> 
> ```js
>     public constructor(settings: OidcClientSettings);
>     public constructor(settings: OidcClientSettingsStore, metadataService: MetadataService); 
>     public constructor(settings: OidcClientSettings | OidcClientSettingsStore, metadataService?: MetadataService) {
>         this.settings = settings instanceof OidcClientSettingsStore ? settings : new OidcClientSettingsStore(settings);
> 
>         this.metadataService = metadataService ?? new MetadataService(this.settings);
> ```
> 
> -> something like
> 
> ```js
> public constructor(settings: UserManagerSettings);
>     public constructor(settings: OidcClientSettings | OidcClientSettingsStore, redirectNavigator?: RedirectNavigator, ...) {
>         this.settings = settings instanceof UserManagerSettingsStore? settings : new UserManagerSettingsStore(settings);
>         ...
>         this._redirectNavigator= redirectNavigator?? new RedirectNavigator(this.settings);
> ```
> 
> That allows you to override the `UserManager` class in your code and inject your navigators too without going to extend the settings for this special cases. And its more like an SDK then...

done

pamapa:
I am  thinking about if it is really needed to expose all navigators as public.
If we would extend the interface `INavigator` with a `callback` function, the types of
```js
    protected readonly _redirectNavigator: RedirectNavigator;
    protected readonly _popupNavigator: PopupNavigator;
    protected readonly _iframeNavigator: IFrameNavigator;
```
could be `INavigator` and we would only need to export that interface `INavigator`...

philjones88:
> I am thinking about if it is really needed to expose all navigators as public. If we would extend the interface `INavigator` with a `callback` function, the types of
> 
> ```js
>     protected readonly _redirectNavigator: RedirectNavigator;
>     protected readonly _popupNavigator: PopupNavigator;
>     protected readonly _iframeNavigator: IFrameNavigator;
> ```
> 
> could be `INavigator` and we would only need to export that interface `INavigator`...

I had considered that too but the 3 navigators have different implementations of callback...

* IFrameNavigator has `callback(url: string, keepOpen = false)`
* PopupNavigator has `callback(url: string)`
* RedirectNavigator doesn't have an implementation

I guess we could do something like:

```
export interface INavigator {
    prepare(params: unknown): Promise<IWindow>;

    callback(url: string, keepOpen?: boolean): Promise<void>;
}
```

the `keepOpen` param doesn't quite sit right with me... but yes it would reduce the exposed surface area by a lot!

pamapa:
this might be even more elegant and easier in the future to extend with additional parameters:
```js
export interface INavigator {
    prepare(params: unknown): Promise<IWindow>;

    callback(url: string, params: { keepOpen?: boolean }): Promise<void>;
}

export class PopupNavigator implements INavigator {
...
    public async callback(url: string, { keepOpen = false }: { keepOpen?: boolean }): Promise<void> {
        this._logger.create("callback");

        PopupWindow.notifyOpener(url, keepOpen);
    }
}
```

pamapa:
we can even hide `keepOpen` in `INavigator` with `unknown`: `callback(url: string, params: unknown): Promise<void>;`

philjones88:
> We can even hide `keepOpen` in `INavigator` with `unknown`: `callback(url: string, params: unknown): Promise<void>;`. I have updated the code above.

Done. The unit tests help ensure it's called correctly. Generally I'm in favour of stronger typings but this seems an ok solution

docs/oidc-client-ts.api.md
```diff
@@ -128,6 +128,16 @@ export interface ILogger {
     warn(...args: unknown[]): void;
 }
 
+// @public (undocumented)
+export interface INavigator {
+    // (undocumented)
+    callback(url: string, params?: unknown): Promise<void>;
+    // Warning: (ae-forgotten-export) The symbol "IWindow" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
philjones88: done

pamapa (approved):

pamapa:
thanks for contributing!

bourquep:
@pamapa Do you think this could be part of a release soon?

pamapa:
i plan to make a release soon, maybe this week or at the beginning of next week.

PR #1158, opened 2023-09-04
Make 'RefreshState' public

### LINKED ISSUE #1157: OIDC Client; can refresh token externally as RefreshState is not exportered
I am using the OIDC client instead of the user manager as need to do some low level stuff with service workers.

But the `useRefreshToken` method of the needs an instance of  RefreshState; which is not exported

davesmits:
Should fix #1157

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -530,6 +530,31 @@ export interface RedirectParams {
     redirectTarget?: "top" | "self";
 }
 
+// @public
+export class RefreshState {
+    constructor(args: {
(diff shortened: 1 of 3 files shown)


pamapa:
change its good, what surprises me is that `docs\oidc-client-ts.api.md` has not changed... ?

pamapa:
> Can you guido me how to get it updated? npm run build doesnt seem to do the trick

I guess you will need to add this type to the index.ts..

pamapa (approved):

pamapa:
thanks for contributing

PR #1173, opened 2023-09-14
Added scriptOrigin parameter for signoutSIlent to work from another origin

### LINKED ISSUE #1172: Allow signoutSilent to work from another origin
I've got a single-spa application with multiple microfrontends that all use a single authentication library that I've wrapped around with oidc-client-ts. We want to use the signoutSilent method in UserManager but it currently does not pass the `scriptOrigin` parameter which prevents us from having the iframe exist on a separate domain.

davidmrobins:
Fixes #1172 

### Checklist


pamapa (approved):

pamapa:
thanks for contributing

PR #1208, opened 2023-10-16
fix: Exception when id_token not set.

### LINKED ISSUE #1206: Invalid token specified: Cannot read properties of undefined (reading 'replace')
I'm testing a service using the Resource Owner Password Credentials flow.  I pass the scope 'openid email profile' , but I don't get an id_token in the response. I don't believe this authorization server supports OIDC for this particular flow.  It just responds with bearer token. 

```
{
    "access_token": "opaque token, not JWT",
    "expires_in": 36000,
    "token_type": "Bearer",
    "scope": "openid profile email",
    "refresh_token": "opaque token, not JWT"
}
```

I was wondering if [this line](https://github.com/authts/oidc-client-ts/blob/f64cfb3d7dd43954c33b771727caa54c863bcec9/src/SigninResponse.ts#L77) should be 
```
return this.scope?.split(" ").includes(OidcScope) && !!this.id_token;
```
instead of 
```
return this.scope?.split(" ").includes(OidcScope) || !!this.id_token;
```

so if an id_token is not present it won't try to parse it at https://github.com/authts/oidc-client-ts/blob/f64cfb3d7dd43954c33b771727caa54c863bcec9/src/ResponseValidator.ts#L54

dopry:
django-oauth-toolkit does not support openid for the resource owner credentials flow. The response will still have openid in the scopes property, but there is not id_token This leads to an exception.

This change makes the client a little more tolerant of auth servers by not trying to parse the id_token in this scenario.

fixes  #1206

### Checklist

- [N/A] This PR makes changes to the public API 
- [*] I have included links for closing relevant issue numbers

### DIFF
--- a/src/ResponseValidator.test.ts
+++ b/src/ResponseValidator.test.ts
@@ -578,6 +578,27 @@ describe("ResponseValidator", () => {
             expect(stubResponse).not.toHaveProperty("profile");
         });
 
+        it("should process an openid signin response without an id_token as a non-openid signin response", async () => {
+            // this is aimed at Authorization servers that don't support the OIDC spec and don't return an id_token if the
+            // openid scope is set in the request scope.
+            // arrange
+            Object.assign(stubResponse, {
+                isOpenId: true,
+            });
+            jest.spyOn(JwtUtils, "decode").mockReturnValue({ sub: "subsub" });
+
+            // act
+            await subject.validateCredentialsResponse(stubResponse, true);
+
+            // assert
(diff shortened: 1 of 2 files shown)


dopry:
@pamapa let me know if you need any thing else on this PR.

pamapa (approved):

pamapa:
thanks for contributing

dopry:
Thanks for maintaining!

PR #1212, opened 2023-10-17
Add support for custom URL state values

### LINKED ISSUE #293: Missing state on Login Redirect
I'm not sure whether this is a bug or I do something wrong, however, I would like to read the state that is preserved between the initial load, redirect to IS and redirect back. 

From the client and IS point of view, everything seems to be fine, the state is being sent all the way through and comes back as expected:

<i>https://<span></span>localhost:44302/connect/authorize?client_id=MyClient&redirect_uri=http://<span></span>localhost:3000/authentication/login-callback&response_type=code&scope=openid profile offline_access&state=**b5924d628ef74c9b8d599a76d7127cd4**&code_challenge=asX_clrphzIXLga3cj7z3gZofVFz-FXxcer25rVsz4k&code_challenge_method=S256&response_mode=query

https://<span></span>localhost:44302/Account/Login?ReturnUrl=/connect/authorize/callback?client_id=MyClient&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fauthentication%2Flogin-callback&response_type=code&scope=openid%20profile%20offline_access&state=**b5924d628ef74c9b8d599a76d7127cd4**&code_challenge=asX_clrphzIXLga3cj7z3gZofVFz-FXxcer25rVsz4k&code_challenge_method=S256&response_mode=query

https://<span></span>localhost:44302/connect/authorize/callback?client_id=MyClient&redirect_uri=http://<span></span>localhost:3000/authentication/login-callback&response_type=code&scope=openid profile offline_access&state=**b5924d628ef74c9b8d599a76d7127cd4**&code_challenge=asX_clrphzIXLga3cj7z3gZofVFz-FXxcer25rVsz4k&code_challenge_method=S256&response_mode=query

http://<span></span>localhost:3000/authentication/login-callback?code=30B4D78B1A42018F8AD3C0579B72A876CFD60936D02998F029BBDBEE3570BED1&scope=openid profile offline_access&state=**b5924d628ef74c9b8d599a76d7127cd4**&session_state=L_rI2q6PV9a_Kljbx5mHL29ftlwEdNQn6k9ky8wak98.3DF5BF2643218A919507CC884F1F1A92</i>

However, if you then look into the storage (I'm using Session storage, but should not make any difference in this case), there is no state with the round-tripped id. I'm using reference tokens and thus I can't read values from tokens as an alternative solution.

Interesting enough, all this works well on sign out redirect. The state id is sent there and back and I can pick up the state from the storage after the final redirect. The issue is not new in `oidc-client-ts`, I've had the same problem with the `oidc-client-js`. 

I'm just wondering whether I do something wrong or whether this is an issue of the library.

jafaircl:
Closes/fixes #293

@pamapa @Badisi 

### Checklist


src/SigninRequest.ts
```diff
@@ -109,7 +111,11 @@ export class SigninRequest {
             parsedUrl.searchParams.append("nonce", nonce);
         }
 
-        parsedUrl.searchParams.append("state", this.state.id);
+        let state = this.state.id;
+        if (url_state) {
+            state = `${state}${URL_STATE_DELIMITER}${url_state}`;
```
Badisi, line 116: Can we actually append more than one value to the parameter ? (ex: `state=ABDC;value1;value2;value3`)
If yes then it means, here, we rely on the developer to use the proper delimiter to separate things inside **url_state**.
I don't know if it's acceptable or not.

jafaircl: I could do what MSAL.js does and join the extra values if you think that's preferrable. The url state is just a string so, as a developer, I would imagine that the string I sent is what I should get back. https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/31fb0a77a6bf9db77a41330fa09a76db4ff030c0/lib/msal-common/src/utils/ProtocolUtils.ts#L104

Badisi: I was just curious to know, but leaving it like this as a string is totally fine 😉
One can even encode a complete object as a string and pass it as the **url_state**.

jafaircl: I think you raise a good point and maybe joining the leftover values is the way to go. MSAL is used by a lot of apps so if they do it, it’s probably not a bad idea.

Yeah any string would work as the state so you could easily stringify an object and pass it. For a moment I thought about just stringifying whatever data they pass using the existing state param. But, it’s possible that it wouldn’t be serializable or would be too big for a url

Badisi: Good catch for the SigninResponse.
This part looks good to me now 😉

pamapa:
@Badisi i let you finish this review, you can tell me when i shall merge this...

pamapa (approved):

PR #1231, opened 2023-11-02
require node >=18

pamapa:
### Checklist


pamapa:
Done with a 3.0.0 a major release and node 18 is maintained where 16 is no longer. I plan to keep that dependency now longer though, probably unless a new major release or demanding feature...

garronej:
Ok, thanks for your answer! However, corect me if I'm mistaking `oidc-client-ts` is a purely client side library that does not comes with any scripts. So ther's nothing really runing in node there is it?

garronej:
SSR I guess, my bad

PR #1233, opened 2023-11-02
improve merge claims behavior

### LINKED ISSUE #1232: improve merge claims behavior
- In order to allow "refresh userinfo data on demand" #877 merging claims must be adapted in way way that we can do it multiple times.
- Changing the type of string to array by default (to preserve the previous different value) seems not a good idea. In case this is needed by some authz servers, we could add a special settings to have this behavior back (see [#881#issuecomment-1425834324](https://github.com/authts/oidc-client-ts/pull/881#issuecomment-1425834324)). `mergeClaimsStrategy?: "replace" | "merge"]`
- drop misleading `mergeClaims` setting, we always merge ...

pamapa:
Closes/fixes #1232

### Checklist

This merge requests enables a fix for #846 and #852. This MR introduces the new merge behavior and moves all the claims related logic inside a dedicated service.
This code is based on the pre-work done by @marcoreni in #881!

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1,6 +1,6 @@
-## oidc-client v2.4.0 &rarr; oidc-client-ts v3.0.0
+## oidc-client-ts v2.4.0 &rarr; oidc-client-ts v3.0.0
 
-The API is largely backwards-compatible.
+The API is largely backwards-compatible. The merge claims behavior has been improved.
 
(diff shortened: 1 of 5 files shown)


pamapa:
@Badisi and @marcoreni Would be awesome if you find time to review this!

PR #1247, opened 2023-11-14
fix: UserSessionChanged should be raised regardless of sid claim. Sid…

### LINKED ISSUE #1242: Session change not fired if identity token does not contain a sid claim
See here: https://github.com/authts/oidc-client-ts/blob/b7266eb650391d9e9807af21164874889963f1be/src/SessionMonitor.ts#L156

If there is no `sid` claim in the id token then `undefined === undefined` evaluates to true and the userSessionChanged event is never fired.

The `sid` claim is not part of the [session management spec](https://openid.net/specs/openid-connect-session-1_0.html), it is only defined in [frontchannel logout spec](https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout) and it's optional. This claim should not be checked at all or at least the user session should be assumed to have changed when the claim is not present.

bachratyg:
… is not part of the session management spec.

Closes/fixes #1242

### Checklist

### DIFF
--- a/src/SessionMonitor.ts
+++ b/src/SessionMonitor.ts
@@ -13,7 +13,6 @@ export class SessionMonitor {
     private readonly _logger = new Logger("SessionMonitor");
 
     private _sub: string | undefined;
-    private _sid: string | undefined;
     private _checkSessionIFrame?: CheckSessionIFrame;
 
     public constructor(private readonly _userManager: UserManager) {
@@ -43,9 +42,8 @@ export class SessionMonitor {
             if (session) {
                 const tmpUser = {
                     session_state: session.session_state,
-                    profile: session.sub && session.sid ? {
+                    profile: session.sub ? {
                         sub: session.sub,
-                        sid: session.sid,
                     } : null,
                 };
                 void this._start(tmpUser);
@@ -56,7 +54,7 @@ export class SessionMonitor {
     protected _start = async (
         user: User | {
             session_state: string;
-            profile: { sub: string; sid: string } | null;
+            profile: { sub: string } | null;
         },
     ): Promise<void> => {
         const session_state = user.session_state;
@@ -67,12 +65,10 @@ export class SessionMonitor {
 
         if (user.profile) {
             this._sub = user.profile.sub;
-            this._sid = user.profile.sid;
             logger.debug("session_state", session_state, ", sub", this._sub);
         }
         else {
             this._sub = undefined;
-            this._sid = undefined;
             logger.debug("session_state", session_state, ", anonymous user");
         }
 
@@ -108,7 +104,6 @@ export class SessionMonitor {
     protected _stop = (): void => {
         const logger = this._logger.create("_stop");
         this._sub = undefined;
-        this._sid = undefined;
 
         if (this._checkSessionIFrame) {
             this._checkSessionIFrame.stop();
@@ -126,9 +121,8 @@ export class SessionMonitor {
                     if (session) {
                         const tmpUser = {
                             session_state: session.session_state,
-                            profile: session.sub && session.sid ? {
+                            profile: session.sub ? {
                                 sub: session.sub,
-                                sid: session.sid,
                             } : null,
                         };
                         void this._start(tmpUser);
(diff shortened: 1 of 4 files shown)


src/SessionMonitor.ts
```diff
@@ -153,13 +153,8 @@ export class SessionMonitor {
                     raiseEvent = false;
                     this._checkSessionIFrame.start(session.session_state);
 
-                    if (session.sid === this._sid) {
```
pamapa, line 156: i do not understand why this should work:
- We still use session.sub at line 152
- With your changes here i would expect that `_raiseUserSessionChanged` is called all the time when the user is still ok...

pamapa: lets keep the `sub` claim

pamapa (approved):

pamapa:
CI pipline fails with:
```
You have changed the public API signature for this project. Please copy the file "temp/oidc-client-ts.api.md" to "docs/oidc-client-ts.api.md", or perform a local build (which does this automatically). See the Git repo documentation for more info.
```

and with
```
npm run lint

> oidc-client-ts@2.4.0 lint

/home/runner/work/oidc-client-ts/oidc-client-ts/src/SessionMonitor.ts
Error:   57:36  error  Unexpected separator (;)  @typescript-eslint/member-delimiter-style
```

please resolve

pamapa:
I just found that `sid` is also part of `SessionStatus` returned in `querySessionStatus`. We should either remove it form there as well or as an alternative we make `sid` optional: If it is defined consider it in the `SessionMonitor` else ignore it. What do you think about the alternative?

bachratyg:
I'd rather not put back a partial `sid` check. That would mean the lib sometimes decides for me if there's a change, and sometimes not. It doesn't feel consistent. This is opinionated behavior, it should be more explicit.

Alternatively the `SessionState` could be extended to include the full user. `querySessionStatus` already does a reduced silent sign-in (only the `openid` scope is requested), might as well do the full thing. The user can also be added to the events then the subscriber can do as they see fit e.g.
```
userManager.events.addUserSessionChanged(async newUser => {
    if( newUser.sid !== oldUser.sid ) {
        // previous handler logic
    }
});
```

pamapa:
> I'd either remove the `sid` or alternatively extend `SessionState` to include the full user.

~~if we are going with the `User`, we could drop `SessionState` and directly use `User`:
-> `querySessionStatus(args: QuerySessionStatusArgs = {}): Promise<User| null>`~~
The returned object would contain the currently signed-in user, would it be useful to also provide the previous signed-in user?

The main question is how do we then handle the types for `monitorAnonymousSession` feature. Do you know the use-case for this feature? If it is still useful, we should add some documentation to the settings property else drop it?

Probably something we already have:
```ts
const tmpUser = {
  session_state: session.session_state,
  profile: session.sub /*&& session.sid*/ ? {
    sub: session.sub,
    //sid: session.sid,
  } : null,
};
```
This would also simplify the type for the parameter of `SessionMonitor._start`.

bachratyg:
> ~if we are going with the `User`, we could drop `SessionState` and directly use `User`: -> `querySessionStatus(args: QuerySessionStatusArgs = {}): Promise<User| null>`~

The `session_state` is still needed for monitoring the anonymous session. It should be exposed separately from `User`.

> The returned object would contain the currently signed-in user, would it be useful to also provide the previous signed-in user?

Subscriber can always call `getUser` but it would certainly be convenient.

> The main question is how do we then handle the types for `monitorAnonymousSession` feature. Do you know the use-case for this feature? If it is still useful, we should add some documentation to the settings property else drop it?

It could be useful for apps that have public features e.g. some sites show a popup when you sign in on another page. The only problem I see here is that a new notification could be triggered while the previous handler is still running but that is already the case with `UserSessionChanged`.

pamapa:
> I'd either remove the `sid` or alternatively extend `SessionState` to include the full user

so lets remove that too...

bachratyg:
Removed `sid` for now. I will need a bit more time to improve querySessionStatus with the full profile

pamapa (approved):

pamapa:
thanks for contributing

PR #1248, opened 2023-11-15
Response mode optional

### LINKED ISSUE #1010: Make Response Mode Optional (Oracle IDCS)
In Oracle IDCS (Identity Cloud Services) both options 
(according to https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html)

- Response Mode for the OAuth 2.0 code Response Type is the query encoding
- Response Mode for the OAuth 2.0 token Response Type is the fragment encoding

are invalid.
If I use 
`query => code` 
I have 
```sh
Oops... Client 111111 provided an invalid response mode: query.
```
in case 
`fragment => token`

```sh
Unhandled Rejection (Error): Only the Authorization Code flow (with PKCE) is supported
```

Only when I remove the default value 

https://github.com/authts/oidc-client-ts/blob/main/src/OidcClient.ts#L102

I can make it work.

In my opinion , if it is not a mistake caused by Oracle developers there are 2 ways to fix.
1 - simply delete the default value
2 - add 'none' option to response_mode

the fix I found out exploring 
https://auth0.com/docs/authenticate/login/oidc-conformant-authentication/oidc-adoption-auth-code-flow#code-exchange-request-authorization-code-flow-with-pkce

pamapa:
Closes/fixes #1010

### Checklist

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -10,6 +10,7 @@ The API is largely backwards-compatible. The merge claims behavior has been impr
   - `refreshTokenCredentials` use `fetchRequestCredentials` since 2.1.0
 - the `mergeClaims` has been replaced by `mergeClaimsStrategy`
   - if the previous behavior is needed `mergeClaimsStrategy: { array: "merge" }` can be used
+- default of `response_mode` changed from `query` &rarr; `undefined`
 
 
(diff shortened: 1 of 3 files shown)


pamapa:
@dzirg44 Does that look good for you?

Issue #971, opened 2023-04-18
How to solve this error: No state in response


LouLamSan:
hi,I am using the latest version of oidc-client.ts, when trying to login and get the token, it prompts me an error: "No state in response". Could you please advise how to solve this issue?

The configuration in my client is as follows:
const config = {
    authority: 'https://localhost:7150/',
    client_id: 'client1',
    redirect_uri: ${window.location.origin}/#/callback,
    response_type: 'code',
    scope: 'openid profile api',
    response_mode: 'query'
}

and the error message like this:
 Error: No state in response
    at OidcClient.readSigninResponseState (OidcClient.ts:155:26)
    at OidcClient.processSigninResponse (OidcClient.ts:173:48)
    at UserManager._signinEnd (UserManager.ts:442:51)
    at UserManager.signinRedirectCallback (UserManager.ts:174:33)
    at Proxy.created (callback.vue:19:31)
    at callWithErrorHandling (runtime-core.esm-bundler.js:173:36)
    at callWithAsyncErrorHandling (runtime-core.esm-bundler.js:182:21)
    at callHook (runtime-core.esm-bundler.js:3608:5)
    at applyOptions (runtime-core.esm-bundler.js:3510:9)
    at finishComponentSetup (runtime-core.esm-bundler.js:7387:9)

the callback url like this:
http://localhost:5007/#/callback?code=9F399AF8C82E4F458A5B1A37C79ADF28&state=b97ddcf6aa584df4a320e3022234bf75

pamapa:
Your response mode is not query but fragment: `response_mode: 'query' -> 'fragment'`

LouLamSan:
That was feasible and the previous error no longer occurs, but now I have received another error message:
Error: authority mismatch on settings vs. signin state
    at ResponseValidator._processSigninState (ResponseValidator.ts:151:26)
    at ResponseValidator.validateSigninResponse (ResponseValidator.ts:62:14)
    at OidcClient.processSigninResponse (OidcClient.ts:175:31)
    at async UserManager._signinEnd (UserManager.ts:442:32)
    at async UserManager.signinRedirectCallback (UserManager.ts:174:22)
    at async Proxy.created (callback.vue:20:13)

I have checked the corresponding source code according to the error message, and I feel confused why the authority that I specified in the configuration is inconsistent with the state.authority（I noticed that this state seems to be read from my configuration.）

pamapa:
[...]

PR #1270, opened 2023-11-26
#971 - add hashRouterMode

tennox:


tennox:
I've just been made aware of this other issue
https://github.com/authts/oidc-client-ts/issues/734#issuecomment-1298481609

And how it's a bug in the IdentityProvider, appending the queryString after the hash part.

So should I cancel this PR?

pamapa:
> So should I cancel this PR?

yes, also see my comment in #971.

tennox:
Got it, closing in favor of https://github.com/pulsejet/nextcloud-oidc-login/issues/251

Thanks :)

arnaudrolly:
#1275 (the doc) seems to have been merged, but not #1270 (the functionality) ?

tennox:
> #1275 (the doc) seems to have been merged, but not #1270 (the functionality) ?

This PR was an attempt at fixing a problem which is not actually real.
Afaik, the behaviour of this lib is correct, and the IdentityProviders I had been using have a bug of putting the query string in the wrong part of the URL.

PR #1300, opened 2023-12-13
explicitly throw an error when crypto.subtle is undefined

### LINKED ISSUE #1299: avoid crypto.subtle is undefined
It is better to throw an exception with a message of the reason.

See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto:
Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto#browser_compatibility).

pamapa:
Closes/fixes #1299

### Checklist

### DIFF
--- a/src/utils/CryptoUtils.ts
+++ b/src/utils/CryptoUtils.ts
@@ -38,6 +38,10 @@ export class CryptoUtils {
      * PKCE: Generate a code challenge
      */
     public static async generateCodeChallenge(code_verifier: string): Promise<string> {
+        if (!crypto.subtle) {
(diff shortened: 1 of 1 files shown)


pamapa:
@43081j Please review this small changes. The idea is to avoid people complaining its undefined when they are using not secure contexts...

43081j (approved):
makes sense to me, good catch!

PR #1342, opened 2024-01-15
Fix too early `UserManager.events().unload()` event before user is actually signed out

### LINKED ISSUE #1341: `UserManager.events().unload()` event is triggered too early on `UserManager.signoutRedirect()`
We observed the `UserManager.events().unload()` event was fired too early, before the user was actually signed out. Our application reacted with a redirect to the login page to this event, but as there was the possibilty of a race condition between login redirect and actual signout, the signout might not yet have happened and therefore resulted in an unsuccessful signout.

The problematic code is located in `UserManager._signoutStart()`, which is called from `UserManager.signoutRedirect()`:

https://github.com/authts/oidc-client-ts/blob/4a8d6d922690077d4edbc86cfe1af363c7b9414c/src/UserManager.ts#L623-L624

I would expect this code instead to be located in `UserManager._signoutEnd()` where it should be guaranteed that the user was successfully signed out.

I will create a PR to fix this issue.

PSanetra:
This PR prevents race conditions with event handlers which react to this unload event and expect the user to be signed out already.

Fixes #1341

### DIFF
--- a/src/UserManager.ts
+++ b/src/UserManager.ts
@@ -620,9 +620,6 @@ export class UserManager {
                 args.id_token_hint = id_token;
             }
 
-            await this.removeUser();
-            logger.debug("user removed, creating signout request");
-
             const signoutRequest = await this._client.createSignoutRequest(args);
             logger.debug("got signout request");
 
@@ -638,10 +635,14 @@ export class UserManager {
             throw err;
         }
     }
+
     protected async _signoutEnd(url: string): Promise<SignoutResponse> {
         const logger = this._logger.create("_signoutEnd");
         const signoutResponse = await this._client.processSignoutResponse(url);
         logger.debug("got signout response");
+
+        await this.removeUser();
+        logger.debug("user removed");
 
         return signoutResponse;
     }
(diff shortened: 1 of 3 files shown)


src/UserManager.ts
```diff
@@ -638,11 +635,20 @@ export class UserManager {
             throw err;
         }
     }
+
     protected async _signoutEnd(url: string): Promise<SignoutResponse> {
         const logger = this._logger.create("_signoutEnd");
         const signoutResponse = await this._client.processSignoutResponse(url);
         logger.debug("got signout response");
 
+        try {
```
pamapa, line 644: please remove this try catch, if `this.removeUser()` there is something really wrong

PSanetra: Done :slightly_smiling_face:

pamapa (approved):

PSanetra:
@pamapa Thank you for reacting so fast! :rocket: 
Is it possible to release a new RC version with this change?

pamapa:
@PSanetra Yes, i can make one in the next few days (1-3)...

Issue #1360, opened 2024-01-25
Demonstrating Proof Of Possession (DPoP)


dbfr3qs:
Recently, the IETF released [RFC9449](https://datatracker.ietf.org/doc/html/rfc9449), a proposed internet standard that describes a method for constraining access tokens to devices through the use of public key cryptography.

The idea is for the client to generate a cryptographic key pair, and use this to sign requests to the authorisation server for access tokens/refresh tokens, and then on subsequent requests to resources with said access tokens/refresh tokens. These signatures are referred to as "proofs" and are supplied in an additional request header. The spec also allows for optionally constraining authorisation codes, so that the entire end to end process is constrained and there's no chance anything can be used outside of the client.

In the frontend, this could be particularly useful because there's no secure method to store tokens, opening up vectors for token exfiltration, for example through XSS or other means (e.g. supply chain, CDN cache pollution). DPoP may provide a way to eliminate certain attack vectors which seek to exfiltrate tokens out of the browser to be used by separate systems owned by bad actors, though it would not protect against attacks where [a compromised browser is used directly by a bad actor](https://www.ietf.org/archive/id/draft-ietf-oauth-browser-based-apps-15.html#section-6.3.3.2.2).

To prevent leaking/exfiltration of the key material used to generate the DPoP proofs in the browser, we can make use of the browser's SubtleCrypto api to generate cryptographic keys that are not extractable. From there, [we can store the CryptoKeyPair object using indexedDB](https://crypto.stackexchange.com/a/52488/113803), and use that object to generate DPoP proofs. This provides a way to securely store key material without the possibility of exfiltration, though I am not sure it would prevent a motivated bad actor who had gain full control of a user's machine or was [proxying requests through a compromised browser](https://www.ietf.org/archive/id/draft-ietf-oauth-browser-based-apps-15.html#name-proxying-requests-via-the-u).

As cookie based authentication becomes more and more restricted by browsers, and refresh tokens become more standard to keep sessions alive, it is my opinion that constraining tokens will become essential, particularly in pure frontend clients.

Is this something that is worth considering adding to the library?

I've provided a [PR](https://github.com/authts/oidc-client-ts/pull/1361) with an implementation that covers most of the spec, and uses the native SubtleCrypto api and indexedDB as described above. I have left out a few areas that I don't think are relevant to the use cases oidc-client-ts addresses. If this is something worth considering, please take a look.

dbfr3qs:
[...]

PR #1361, opened 2024-01-25
Demonstrating Proof of Possession

dbfr3qs:
Addresses/highlights #1360 

As per the issue, this is an example of a potential implementation of Demonstrating Proof of Possession (DPoP) to constrain access/refresh tokens to devices. I'm keen on feedback/collaboration if this is a thing that is deemed worth adding to the library...

As mentioned in the issue, this is mostly spec complete - though I have left out some parts that I don't think are appropriate to the oidc-client-ts use case (such as [Pushed Authorisation Requests](https://datatracker.ietf.org/doc/html/rfc9449#name-dpop-with-pushed-authorizat)).

I've used the [react-odic-client library example app](https://github.com/dbfr3qs/react-oidc-context/tree/DPoP-oidc-client-ts) and npm linked it to my local version of this PR. I used a fork of [IdentityServer's DPoP examples](https://github.com/dbfr3qs/Samples/tree/DPoP-oidc-client-ts) to develop and test against, if you are interested in trying it out for yourself.

**Things of note**
- This adds a new set of options specifically for DPoP to the OidcClient and introduces a new service - the DPoPService. 
- What's the general feeling in regards to adding dependencies? I ask this because...
    - I've included two dependencies to do this: [Jose](https://github.com/panva/jose) and [idb-keyval](https://github.com/jakearchibald/idb-keyval). Jose allows for generating JWTs at runtime and idk-keyval wraps the native IndexedDB api with a more intuitive promise based api. Both of these libraries have no external dependencies.
    - I had to get a little bit funky in order to satisfy these dependencies in the tests.
    - It will increase the size of the overall bundle.
- In the case of [crypto-js](https://github.com/brix/crypto-js#discontinued), I see that it is no longer maintained. I've used the browser native [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey) api to generate key material here in any case, because I wanted to take advantage of the [extractable](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/extractable) property when generating key pairs.

I'm not super experienced with Typescript, apologies if I've made rookie mistakes. Please give me feedback.

### Checklist


src/JsonService.ts
```diff
@@ -176,6 +187,15 @@ export class JsonService {
         if (!response.ok) {
             logger.error("Error from server:", json);
             if (json.error) {
+                if (json.error === "use_dpop_nonce" && dpopEnabled && response.headers.get("Dpop-Nonce")) {
```
dbfr3qs, line 190: The spec [describes an optional nonce requirement](https://datatracker.ietf.org/doc/html/rfc9449#name-resource-server-provided-no) whereby if a nonce is not supplied during a request, a bad request is returned by the resource server with a specific response body and the client is expected to generate a nonce and retry the request.

src/DPoPService.ts
```diff
@@ -0,0 +1,102 @@
...
+            } else {
+                throw err;
+            }
+        }
+    }
+
+    protected static async generateKeys() : Promise<CryptoKeyPair> {
+        return await window.crypto.subtle.generateKey(
+            {
+                name: "ECDSA",
```
dbfr3qs, line 95: The spec only goes as far as recommending that signing algorithms are asymmetric:
>Implementers MUST ensure that only asymmetric digital signature algorithms (such as `ES256`) that are deemed secure can be used for signing DPoP proofs. In particular, the algorithm `none` MUST NOT be allowed

I've hardcoded to ES256 for now, as it is highly recommended by the [IANA](https://www.iana.org/assignments/jose/jose.xhtml#:~:text=ES384,n/a).

src/User.ts
```diff
@@ -124,4 +125,8 @@ export class User {
         Logger.createStatic("User", "fromStorageString");
         return new User(JSON.parse(storageString));
     }
+
+    public async dpopProof(url: string, httpMethod?: string): Promise<string> {
```
dbfr3qs, line 129: It feels a bit weird exposing this here: there needs to be a way of allowing consumers to generate a DPoP proof and then attaching that to authenticated requests.

But it seemed like the easiest way from the pov of the react-oidc-client library... for example https://github.com/dbfr3qs/react-oidc-context/blob/217d6d18109fc82526d7abc4015a98ec359fe79b/example/index.tsx#L29C13-L39C16

jest-environment-jsdom.cjs
```diff
@@ -0,0 +1,28 @@
...
+
+const { TextEncoder, TextDecoder } = require('util');
+const { default: $JSDOMEnvironment, TestEnvironment } = require('jest-environment-jsdom');
+const crypto = require("crypto");
+
+Object.defineProperty(exports, '__esModule', {
+    value: true
+});
+
+class JSDOMEnvironment extends $JSDOMEnvironment {
```
dbfr3qs, line 11: jsdom doesn't support some apis so I had to work around that.
https://github.com/jsdom/jsdom/issues/2524

Also some issues with jest and esm modules...
https://stackoverflow.com/questions/76608600/jest-tests-are-failing-because-of-an-unknown-unexpected-token-export

src/JsonService.ts
```diff
@@ -131,12 +133,21 @@ export class JsonService {
...
+        dpopEnabled,
     }: PostFormOpts): Promise<Record<string, unknown>> {
         const logger = this._logger.create("postForm");
+        let DPoPProof;
+
         const headers: HeadersInit = {
             "Accept": this._contentTypes.join(", "),
             "Content-Type": "application/x-www-form-urlencoded",
         };
+
```
pamapa, line 145: instead of limiting this to DPoP header, we could simply enrich the interface with a generic: `extraHeaders?: Record<string, ExtraHeader>;` parameter...

src/JsonService.ts
pamapa: lets keep the throw of that error, if the IDP needs a `Dpop-Nonce`, we should already provide one in the first place...

dbfr3qs: Yeah, I'm okay with that but it's worth considering that the spec is a little unclear about when an authorization server should require a dpop_nonce (keep in mind that the nonce is generated server side):

>Including a nonce value contributed by the authorization server in the DPoP proof MAY be used by authorization servers to limit the lifetime of DPoP proofs. The server determines when to issue a new DPoP nonce challenge and if it is needed, thereby requiring the use of the nonce value in subsequent DPoP proofs. The logic through which the server makes that determination is out of scope of this document.

but it's quite clear about what to do when no nonce is supplied:

>An authorization server MAY supply a nonce value to be included by the client in DPoP proofs sent. In this case, the authorization server responds to requests that do not include a nonce with an HTTP 400 (Bad Request) error response per [Section 5.2](https://rfc-editor.org/rfc/rfc6749#section-5.2) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] using use_dpop_nonce as the error code value. The authorization server includes a DPoP-Nonce HTTP header in the response supplying a nonce value to be used when sending the subsequent request. Nonce values MUST be unpredictable. This same error code is used when supplying a new nonce value when there was a nonce mismatch. The client will typically retry the request with the new nonce value supplied upon receiving a use_dpop_nonce error with an accompanying nonce value.

This is how it's implemented in IdentityServer for example - the nonce header isn't returned anywhere in the flow, prior to responding with a 400 bad request if the drop_nonce header doesn't exist. This is also how [Okta](https://developer.okta.com/docs/guides/dpop/main/#oauth-2-0-dpop-jwt-flow) does it. So, happy to remove it, but I suspect what is here will become the norm if it hasn't already.

src/DPoPService.ts
```diff
@@ -0,0 +1,102 @@
+import { base64url, exportJWK, SignJWT, calculateJwkThumbprint } from "jose";
```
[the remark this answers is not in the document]
dbfr3qs: No problem. I originally tried to implement the JWT creation code using the native subtle crypto library but ran into trouble with some of the encoding/decoding before switching to a library that I knew would just work. I can try again though.

On that note, I have also included the `idv-keyval` package, which itself has no dependencies, to handle the indexedDB queries. I've found that using the native IndexedDB API is... challenging. idb-keyval is very small and has no dependencies itself, but what do you think?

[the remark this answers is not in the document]
dbfr3qs: I have removed Jose as a run time dependency, but I have left it as a dev dependency to use in tests (to verify jwts, for example). Is this okay?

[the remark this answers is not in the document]
pamapa: nice that you have removed the dependency in production code, for unit-test it is ok to have that dependency.

src/UserManager.ts
```diff
@@ -15,6 +15,7 @@ import type { SignoutResponse } from "./SignoutResponse";
 import type { MetadataService } from "./MetadataService";
 import { RefreshState } from "./RefreshState";
 import type { SigninResponse } from "./SigninResponse";
+import { del } from "idb-keyval";
```
pamapa, line 18: The `idb-keyval` must be replaced with the existing store user infrastrcuture. There where we store the user/session we can store the `dpop` value too...

dbfr3qs: No worries, I am working on removing that dependency now.

However, we must use indexedDB to store non extractable crypto key objects. We cannot store these in a plain text storage mechanism, such as local/session storage. If we do this, it provides some measure of protection against leaking the crypto key material used to generate the DPoP proofs. If we chose to make the key material extractable and stored in plane text, then we largely defeat the purpose of DPoP in the browser as the DPoP keys can be exfiltrated out of the browser alongside any tokens.

See
https://stackoverflow.com/a/49479890
https://crypto.stackexchange.com/a/52488
https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/extractable

And the ietf recommendation https://www.ietf.org/archive/id/draft-ietf-oauth-browser-based-apps-16.html#section-6.3.3.2.2

Instead I will provide a similar interface to the WebStorageStageStore, though wrapping IndexedDB and specifically for storing CryptoKeyPairs. This will remove the dependency on `ida-keyval`, but the key material will still be stored in IndexedDB.

Does this make sense?

src/JsonService.ts
dbfr3qs: Hey @pamapa I have removed the authorization server supplied nonce support for the time being. I think it would be better to do it in a separate PR if that's okay (it is [optional](https://datatracker.ietf.org/doc/html/rfc9449#section-8) in the spec)? This PR is large and it was becoming messy.

dbfr3qs:
@pamapa I think this is ready for a proper review as it is essentially code complete, apart from optional AS supplied nonce support which I think should be done separately at a later date. Please take a look and let me know what else you'd like to change, or any questions you may have as I know it's quite a large PR.

pamapa:
> Please take a look and let me know what else you'd like to change, or any questions you may have as I know it's quite a large PR.

It is still very invasive and i am not convinced all new code is at the correct place.
- e.g. `TokenClient.ts` does not need `extraHeaders["DPoP"] = ...`, but simply support the `extraHeaders` parameter too

At the end probably (i am not 100% sure) all/most code adaptions, which are `DPoP` related end up in the `UserManager` class.

Maybe it makes sense in a first step to only extend this library in a generic way (like you did with `extraHeaders`), such that you can inherit e.g. the `UserManager` class yourself and are able to use this feature.

dbfr3qs:
> > Please take a look and let me know what else you'd like to change, or any questions you may have as I know it's quite a large PR.
> 
> It is still very invasive and i am not convinced all new code is at the correct place.
> 
> * e.g. `TokenClient.ts` does not need `extraHeaders["DPoP"] = ...`, but simply support the `extraHeaders` parameter too
> 
> At the end probably (i am not 100% sure) all/most code adaptions, which are `DPoP` related end up in the `UserManager` class.
> 
> Maybe it makes sense in a first step to only extend this library in a generic way (like you did with `extraHeaders`), such that you can inherit e.g. the `UserManager` class yourself and are able to use this feature.

No problems - that all makes sense 👍

dbfr3qs:
@pamapa is this more in line with what you were thinking?

I have pulled everything up into the UserManager. DPoP related functionality is then passed through as either an extraHeader (the `DPoP` header) or as a parameter (`dpopJkt` which is added as a url param during sign in to bind the DPoP proof to the authorization code).

src/OidcClient.ts
```diff
@@ -24,6 +24,7 @@ export interface CreateSigninRequestArgs
     redirect_uri?: string;
     response_type?: string;
     scope?: string;
+    dpopJkt? : string;
```
pamapa, line 27: i guess this can be part of extra header you add just below

pamapa:
The changes go in the right direction, thanks for being patience.

I guess it makes sense to have a separate merge request with stuff you need, but does not contain DPoP, this way you have progress and its easier to review:
- JsonService.ts
- TokenClient.ts
- ResponseValidator.ts
- typos : `recommend` -> `recommended`

dbfr3qs:
> The changes go in the right direction, thanks for being patience.
> 
> I guess it makes sense to have a separate merge request with stuff you need, but does not contain DPoP, this way you have progress and its easier to review:
> 
> * JsonService.ts
> * TokenClient.ts
> * ResponseValidator.ts
> * typos : `recommend` -> `recommended`

That's a great idea - I can do that.

src/OidcClient.ts
dbfr3qs: I don't think so - it's not sent as a header, it's a url request parameter in the sign up request (to /authorize) so that the AS has a way of knowing that the subsequent /token request comes from the same entity.

The extraHeaders, at this point, are exclusively used to send along the DPoP proof which is used when making said subsequent requests for tokens.

Unless I'm misunderstanding you?

dbfr3qs:
Closing this in favour of #1461

PR #1383, opened 2024-02-05
signoutRedirect not working

### LINKED ISSUE #1379: UserManager signoutRedirect not working
To logout I use the[ signoutRedirect method of the UserManager](https://authts.github.io/oidc-client-ts/classes/UserManager.html#signoutRedirect).
That worked perfectly fine with version 2.4.0.

With version 3.0.0 the user is no longer properly logged out.

oidc-client-ts correctly redirects the user agent to the auth server with id_token_hint and post_logout_redirect_uri.
The auth server performs the logout as usual and redirects the user agent back to the application.

The difference compared to version 2.4.0 is, that the UserManager still holds a user object, accesible by the [getUser method](https://authts.github.io/oidc-client-ts/classes/UserManager.html#getUser).

Causing the issue should be the removal of [this line in the _signoutStart method](https://github.com/authts/oidc-client-ts/blob/v2.4.0/src/UserManager.ts#L552).

Is this a bug or should I change some things in my application?

pamapa:
Closes/fixes #1379

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -989,7 +989,7 @@ export class UserManager {
     protected _signout(args: CreateSignoutRequestArgs, handle: IWindow): Promise<SignoutResponse>;
     signoutCallback(url?: string, keepOpen?: boolean): Promise<void>;
     // (undocumented)
-    protected _signoutEnd(url: string): Promise<SignoutResponse>;
+    protected _signoutEnd(url: string, removeUser: boolean): Promise<SignoutResponse>;
     signoutPopup(args?: SignoutPopupArgs): Promise<void>;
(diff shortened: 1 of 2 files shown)


pamapa:
@PSanetra and @ch-lepp Something like this? I have not yet adapted the unit-tests...

PSanetra:
@pamapa sorry, I don't have much time today to look into this. I can look into this tomorrow. But please consider adding a setting to UserManager to configure if a signout redirect is expected or not. (Like mentioned in #1379)

pamapa:
This does not solve this case:
- the callback might not be called by the IDP (e.g. no valid session), the client stays in a non fixable state (the user is still stored locally and will not be removed due to the not coming callback)

PR #1391, opened 2024-02-07
add setting to configure when to raise the user unload event and remove user always before requesting sign-out

### LINKED ISSUE #1379: UserManager signoutRedirect not working

pamapa:
Closes/fixes #1379

### Checklist

### DIFF
--- a/src/UserManager.ts
+++ b/src/UserManager.ts
@@ -151,10 +151,14 @@ export class UserManager {
      * @returns A promise
      */
     public async removeUser(): Promise<void> {
-        const logger = this._logger.create("removeUser");
+        await this._removeUser(true);
+    }
+
+    protected async _removeUser(raiseEvent: boolean): Promise<void> {
+        const logger = this._logger.create("_removeUser");
         await this.storeUser(null);
         logger.info("user removed from storage");
-        await this._events.unload();
+        await this._events.unload(raiseEvent);
     }
 
     /**
@@ -528,6 +532,11 @@ export class UserManager {
      */
     public async signoutRedirect(args: SignoutRedirectArgs = {}): Promise<void> {
         const logger = this._logger.create("signoutRedirect");
+
+        if (!this.settings.raiserUserUnloadEventBeforeSignoutRequest && !this.settings.post_logout_redirect_uri) {
+            throw new Error("post_logout_redirect_uri"); // to raise unload event
+        }
+
         const {
             redirectMethod,
             ...requestArgs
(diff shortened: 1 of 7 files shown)


pamapa:
@PSanetra Can you review and test?

- unload event is
  - deferred to callback (default, `raiserUserUnloadEventBeforeSignoutRequest=false`)
  - done before requesting the sign-out  (option: `raiserUserUnloadEventBeforeSignoutRequest=true`)
- remove user from storage is always (independent of option) done before requesting the sign-out

Any better name for `raiserUserUnloadEventBeforeSignoutRequest`?

src/UserManager.ts
```diff
@@ -641,8 +654,9 @@ export class UserManager {
         const signoutResponse = await this._client.processSignoutResponse(url);
         logger.debug("got signout response");
 
-        await this.removeUser();
-        logger.debug("user removed");
+        if (!this.settings.raiserUserUnloadEventBeforeSignoutRequest) {
```
pamapa, line 657: maybe we should just call `await this._removeUser(true)` here, it should not hurt and is simpler (less code)

PSanetra: I think this should be ok as a signed in user would be unexpected in any case, right?

PSanetra:
I think `raiserUserUnloadEventBeforeSignoutRequest` might be renamed to `noSignoutCallbacksSupported` or `signoutCallbackSupported` with `true` as default if the value is `undefined`.

The user/ developer is expected to set `signoutCallbackSupported` to false if he expects never to call any signout callback.

pamapa:
I think this solution is too complicated, i do not like an option for this at all. The previous behavior (<v3.0) was dead simple

PR #1405, opened 2024-02-15
enable using default scopes from authorization server

natergj:
OAuth allows for the authorization server to provide default scopes when no scopes are specified by the requestor as described in https://datatracker.ietf.org/doc/html/rfc6749#section-3.3

This allows for the library to be configured to rely on the default scopes.

Closes/fixes # n/a

### Checklist

- [x] N/A I have included links for closing relevant issue numbers


src/OidcClientSettings.ts
```diff
@@ -137,6 +137,12 @@ export interface OidcClientSettings {
      * Only scopes in this list will be passed in the token refresh request.
      */
     refreshTokenAllowedScope?: string | undefined;
+
+    /**
+     * https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 describes behavior when omitting scopes from sign in requests
+     * If the IDP supports default scopes, this setting will ignore the scopes property passed to the config
+     */
+    useIdpDefaultScopes?: boolean;
```
natergj, line 145: the current behavior when not specifying a scope value is to default the value to "openid". Changing that would be a breaking change for the library. 

This field was added to prevent a breaking change.  

If the maintainers are OK with a breaking change, I think it would be more straightforward to remove the default value for scope instead of adding this new property. I can set up a PR for that approach if it is desired over adding this property.

pamapa: Good point, lets keep this settings property for now. This give us time to decide when to do the breaking change. I will have to make a v4.0.0 for this change...

Can you immediately deprecate the new property?

natergj: sure thing.  The deprecated tag has been added and pushed. Thanks @pamapa 

I was looking more into this implementation. I think it would be awesome if the library supported this feature, but I realized that the doc I linked to in the description is the part of the [OAuth spec file](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3). The [OIDC authentication request spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) does state that the `openid` scope is required for a valid authentication request.

The OIDC specs  does also state
> "If the openid scope value is not present, the behavior is entirely unspecified." 

If it would make it more clear, I could also rename the property to `allowOidcScopeOmission` or something similar.

pamapa: > The [OIDC authentication request spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) does state that the `openid` scope is required for a valid authentication request.

Oh yes, that makes the `@deprecated` invalid :-) Also its better to keep the type of scope `string`. I guess you can deal with empty string...

According to https://datatracker.ietf.org/doc/html/rfc6749#section-3.3:
"If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined)."
the new property could be named `omitScopeWhenRequesting`?

What kind of Idp are you using for this feature?

src/OidcClientSettings.ts
```diff
@@ -157,10 +165,12 @@ export class OidcClientSettingsStore {
     public readonly client_id: string;
     public readonly client_secret: string | undefined;
     public readonly response_type: string;
-    public readonly scope: string;
+    public readonly scope: string | undefined;
     public readonly redirect_uri: string;
     public readonly post_logout_redirect_uri: string | undefined;
-    public readonly client_authentication: "client_secret_basic" | "client_secret_post";
+    public readonly client_authentication:
```
pamapa, line 171: I have seen multiple unnecessary and not related code style changes, please remove them...

natergj: doh. forgot to turn off the auto-formatter on save. These should be fixed now.

src/OidcClientSettings.ts
natergj: We're using ForgeRock (now part of Ping Identity) as our IDP. We do have the ability to define the IDP behavior when scope is not provided. 

I've removed the updates to the scope typing so it is no longer optional where it was not optional before. The new property has been renamed `omitScopeWhenRequesting` and that property value is solely responsible for including the `scope` query param in the authorization requests now.   The overall changes in this PR are much smaller now. Thanks for all the feedback and suggestions.

natergj:
@pamapa looks like this won't be part of the 3.1.0 release? Are there any updates or anything else I can do on this PR to help get it merged?

pamapa:
> @pamapa looks like this won't be part of the 3.1.0 release? Are there any updates or anything else I can do on this PR to help get it merged?

Good that you ask. Do you have an actual use-case why you would need such a feature and let us know? You will also need to rebase against main, as it as now merge conflicts.

src/OidcClient.ts
```diff
@@ -136,6 +137,7 @@ export class OidcClient {
             skipUserInfo,
             nonce,
             disablePKCE: this.settings.disablePKCE,
+            omitScopeWhenRequesting: omitScopeWhenRequesting,
```
pamapa, line 140: here we can use a shorthand: `omitScopeWhenRequesting: omitScopeWhenRequesting,` -> `omitScopeWhenRequesting ,`

natergj:
> Good that you ask. Do you have an actual use-case why you would need such a feature and let us know? You will also need to rebase against main, as it as now merge conflicts.

Ah. missed the merge conflicts. Those are resolved now. 

there are three use cases we are interested in for the default scopes

1. The issue we are starting to run into with some of our clients is request urls are too large due to a very long scope search param for scopes being requested and the requests were being rejected. The APIs our web apps interact with use very verbose scopes and we need to request a lot of them so our query string becomes very long. By using default scopes, our requests to the IDP will be reduced in size significantly without needing to make additional adjustments to our hosting infrastructure. 

2. Our web application teams have been provided with a process to request new allowed scopes for their web applications. The web application must then be updates to request the newly allowed scope after it is added to the client. If a web app begins requesting the scope prior to it being approved, all subsequent authentication requests fails and a deployment must be rolled back. It is our hope that the default scope will help to alleviate the human error as the web apps will not become out of sync with their requested scopes.

3. Our mobile app uses a graphql interface. Our API teams are updating the REST APIs that back the graph. These updates will often require a new or different scope to be included in the authorization token. Default scopes will allow that work to continue without the need to deprecate and replace the graphql schema or forcing our users to continuously be updating the mobile app.

src/OidcClientSettings.ts
```diff
@@ -142,6 +142,12 @@ export interface OidcClientSettings {
      * Defines request timeouts globally across all requests made to the authorisation server
      */
     requestTimeoutInSeconds?: number | undefined;
+
+    /**
+     * https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 describes behavior when omitting scopes from sign in requests
+     * If the IDP supports default scopes, this setting will ignore the scopes property passed to the config
```
pamapa, line 148: please add: (Default: false) at the end of the comment, see other comments...

src/OidcClient.ts
natergj: updated

src/OidcClientSettings.ts
natergj: updated

pamapa (approved):

natergj:
Thanks for your time and help with this @pamapa ! It's greatly appreciated

pamapa:
Thanks for contributing!

PR #1430, opened 2024-03-06
Add global requestTimeoutInSeconds setting to OidcClientSettings

### LINKED ISSUE #1414: Add support to define JsonService timeout
Given the following use case:
1. App starts
2. User is not logged in
3. A signin silent with iframe is triggered (and awaited because if it fails, user won't be allowed to go further)
4. Down the road a [call](https://github.com/authts/oidc-client-ts/blob/main/src/OidcClient.ts#L119) to `metadataService._getMetadataProperty("authorization_endpoint")` is made
5. Then a [call](https://github.com/authts/oidc-client-ts/blob/main/src/JsonService.ts#L95) to `jsonService.fetchWithTimeout()`

From there, if the identity provider is not reachable, the `fetch` call will timed out only after 90s (depends on web browser implementation). But because 3. was awaited, the app is also frozen for 90s - nothing happens, maybe a white screen, and user is left in the unknown..

In such a case it would be nice to be able to define the timeout ourselves, to reduce the awaiting time.
(something like `silentRequestTimeoutInSeconds`)

dbfr3qs:
Closes #1414 

This adds a `requestTimeoutInSeconds` setting. When this is defined, it applies a retry timeout to all fetch requests made to the authorization server (e.g. /token, /userinfo, /.well-known/openid-configuration) except for requests made with  `silentRequestTimeoutInSeconds` (if it is defined). 

So the logic goes

1.  requestTimeoutInSeconds is set -> all requests have the same timeout
2. silentRequestTimeoutInSeconds is set -> only silent request have a specific timeout
3. both are set -> all requests have the same timeout except silent requests are set to silentRequestTimeoutInSeconds

### Checklist


src/OidcClientSettings.ts
```diff
@@ -201,6 +207,7 @@ export class OidcClientSettingsStore {
         // behavior flags
         filterProtocolClaims = true,
         loadUserInfo = false,
+        requestTimeoutInSeconds ,
```
pamapa, line 210: please remove the extra space before the ,

pamapa:
@Badisi Can you have a look at this?

src/UserManager.ts
```diff
@@ -332,7 +332,7 @@ export class UserManager {
     protected async _useRefreshToken(args: UseRefreshTokenArgs): Promise<User> {
         const response = await this._client.useRefreshToken({
             ...args,
-            timeoutInSeconds: this.settings.silentRequestTimeoutInSeconds,
+            timeoutInSeconds: this.settings.requestTimeoutInSeconds ?? this.settings.silentRequestTimeoutInSeconds,
```
pamapa, line 335: There are more places where `silentRequestTimeoutInSeconds` (e.g. `IFrameNavigator`) is used. I guess it would be better to handle that in `UserManagerSettingsStore`:

Idea if requestTimeoutInSeconds  is set we use the same value for `silentRequestTimeoutInSeconds`

something like
```
this.silentRequestTimeoutInSeconds = requestTimeoutInSeconds ?? silentRequestTimeoutInSeconds,
```

Badisi:
> @Badisi Can you have a look at this?

I'm abroad right now, will be able to look into this in 2 weeks

dbfr3qs:
@pamapa @Badisi Thankyou both for your excellent feedback. I've made the suggested changes to try and satisfy the rules described above, such that the `silentRequestTimeoutInSeconds` setting is not overridden by `requestTimeoutInSeconds` unless it is not defined:

> Either:
> 
> 1. `requestTimeoutInSeconds` is set -> all requests have the same timeout
> 2. `silentRequestTimeoutInSeconds` is set -> only silent request have a specific timeout
> 3. both are set -> all requests have the same timeout except silent request that have a different one

However, there's a default setting for the `silentRequestTimeoutInSeconds` setting so the logic will likely not work unless `silentRequestTimeoutInSeconds` is intentionally set to undefined. Thoughts?

pamapa:
> I would go with the opposite: `silentRequestTimeoutInSeconds` should override `requestTimeoutInSeconds` if it is also set. Otherwise I wouldn't be able to define a global timeout + a different one for silent request.

Makes sense! We can handle this also within `UserManagerSettingsStore`...

dbfr3qs:
> @pamapa @Badisi Thankyou both for your excellent feedback. I've made the suggested changes to try and satisfy the rules described above, such that the `silentRequestTimeoutInSeconds` setting is not overridden by `requestTimeoutInSeconds` unless it is not defined:
> 
> > Either:
> > 
> > 1. `requestTimeoutInSeconds` is set -> all requests have the same timeout
> > 2. `silentRequestTimeoutInSeconds` is set -> only silent request have a specific timeout
> > 3. both are set -> all requests have the same timeout except silent request that have a different one
> 
> However, there's a default setting for the `silentRequestTimeoutInSeconds` setting so the logic will likely not work unless `silentRequestTimeoutInSeconds` is intentionally set to undefined. Thoughts?

Any thoughts on this @Badisi @pamapa ?

pamapa:
> However, there's a default setting for the `silentRequestTimeoutInSeconds` setting so the logic will likely not work unless `silentRequestTimeoutInSeconds` is intentionally set to undefined. Thoughts?

True, `silentRequestTimeoutInSeconds` is set by default to 10s. In order to not break the existing behavior we can not leave it undefined.

Probably something like this within `UserManagerSettingsStore` should work:
```ts
 this.requestTimeoutInSeconds = requestTimeoutInSeconds;
 this.silentRequestTimeoutInSeconds = silentRequestTimeoutInSeconds ? silentRequestTimeoutInSeconds
   : requestTimeoutInSeconds ? requestTimeoutInSeconds
   : DefaultSilentRequestTimeoutInSeconds;
```

dbfr3qs:
Let me know if this does the trick @pamapa @Badisi 🙏

pamapa:
Looks good to me. @Badisi What do you think?

pamapa (approved):

pamapa:
Thanks for contributing and being patient!

PR #1457, opened 2024-04-02
Add extra headers

dbfr3qs:
This adds the extraHeaders plumbing that will be required to implement DPoP, re: #1361 as per [this request](https://github.com/authts/oidc-client-ts/pull/1361#issuecomment-2011905006).

Decided not to both adding extraHeaders to methods relating to the Resource Owner Password Credentials grant as it Is actively discouraged in the latest version of the [OAuth2 BCP](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-resource-owner-password-cre).

### Checklist


pamapa (approved):

pamapa:
thanks for contributing

PR #1461, opened 2024-04-03
Add demonstrating proof of possession

dbfr3qs:
This is the second part of splitting up #1361 - hopefully it is easier to review.

Addresses/highlights https://github.com/authts/oidc-client-ts/issues/1360

As per the issue, this is an implementation of Demonstrating Proof of Possession (DPoP) to constrain access/refresh tokens to devices.

This is mostly spec complete - though I have left out some parts that I don't think are appropriate to the oidc-client-ts use case (such as [Pushed Authorisation Requests](https://datatracker.ietf.org/doc/html/rfc9449#name-dpop-with-pushed-authorizat)). I have also left out support for Authorization Server supplied nonces, which I would like to address in a separate PR in the future (this is optional in the spec).

I've used the [react-odic-client library example app](https://github.com/dbfr3qs/react-oidc-context/tree/DPoP-oidc-client-ts) and npm linked it to my local version of this PR. I used a fork of [IdentityServer's DPoP examples](https://github.com/dbfr3qs/Samples/tree/DPoP-oidc-client-ts) to develop and test against, if you are interested in trying it out for yourself.

Things of note:

- This adds a new set of options specifically for DPoP to the UserManagerSettings and introduces a new service - the DPoPService. This is off by default.
- It makes use of IndexedDb, which allows storing [non extractable key material](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey/extractable) used to generate the DPoPProofs. This means we can still use the stored keys to generate DPoP proofs, even though the key values themselves can't be accessed by the browser directly. This is critical to this functionality and without it DPoP is pointless in the browser context as the keys can just be exfiltrated by a bad actor alongside any tokens if stored in plain text. Storing key material this way prevents an attack where a bad actor may be able to exfiltrate tokens, because the actor cannot generate the required proof outside of the browser/session context.

### Checklist


dbfr3qs:
Will work on getting the code coverage up over the next week or so.

src/UserManager.ts
```diff
@@ -330,11 +352,10 @@ export class UserManager {
         return user;
     }
 
-    protected async _useRefreshToken(args: UseRefreshTokenArgs, extraHeaders?: Record<string, ExtraHeader>): Promise<User> {
+    protected async _useRefreshToken(args: UseRefreshTokenArgs): Promise<User> {
         const response = await this._client.useRefreshToken({
             ...args,
             timeoutInSeconds: this.settings.silentRequestTimeoutInSeconds,
-            extraHeaders,
```
pamapa, line 337: why the remove of `extraHeaders`

dbfr3qs: It get's passed in as part of the UseRefreshTokenArgs - I think it was initially a mistake.

pamapa: makes sense, maybe you can send a separate MR for this to fix?

dbfr3qs: No problem - will do 👍

pamapa:
I am thinking about if it makes sense to add an extended userStore for dpop. Within that store we could handle/ isolate stuff like `await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");`. We might even merge this new `userStore ` class with `IndexedDbCryptoKeyPairStore`...

````ts
export class DPopStorageStateStore extends WebStorageStateStore {
  // ... handles indexeddb stuff and dpop related things needed to be done differently than without dpop
}
```

dbfr3qs:
> I am thinking about if it makes sense to add an extended userStore for dpop. Within that store we could handle/ isolate stuff like `await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");`. We might even merge this new `userStore ` class with `IndexedDbCryptoKeyPairStore`...
> 
> ```ts
> export class DPopStorageStateStore extends WebStorageStateStore {
>   // ... handles indexeddb stuff and dpop related things needed to be done differently than without dpop
> }
> ```
> ```

I'll give it a try and see what I come up with 👍

dbfr3qs:
> I am thinking about if it makes sense to add an extended userStore for dpop. Within that store we could handle/ isolate stuff like `await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");`. We might even merge this new `userStore ` class with `IndexedDbCryptoKeyPairStore`...
> 
> ```ts
> export class DPopStorageStateStore extends WebStorageStateStore {
>   // ... handles indexeddb stuff and dpop related things needed to be done differently than without dpop
> }
> ```
> ```

Okay, I had a Quick Look...

When you say extend the WebStorageStateStore class, what benefit do we get from this? 

Because we are dealing with the CryptoKeyPair type when dealing with IndexedDb (this is the type required to be able to store the key material as non extractable and also the type/supertype used by the crypto.subtle library), we can't override the set, get, remove and getAllKeys methods on the class because the StateStore interface specifies the types <string | null>.

So adding additional methods, such as getDPoPKeyPair is fine, but we don't really need the methods set, get, etc that are exposed by WebStorageStateStore because we don't need the session/local storage for anything DPoP currently (we might do if/when we add support for server side nonces though).

We could change the StateStore interface to allow for CyptoKeyPair, e.g. something like:

```
export interface StateStore<T extends string | null | CryptoKeyPair> {
    set(key: string, value: T): Promise<void>;
    get(key: string): Promise<T>;
    remove(key: string): Promise<T>;
    getAllKeys(): Promise<string[]>;
}
```

And implement that interface directly e.g. `export class DPoPStorageStateStore implements StateStore<CryptoKeyPair | null>` but that would mean making slight changes to the classes that already implement it as well I think (to specify the types). If we did this and decided to add the ability to also store DPoP nonces as strings in local/session storage through the same class, we could then widen the accepted types to <string | null | CryptoKeyPair> and add logic to put nonces in local/session storage and key material in IndexedDB. 

Anyway, I suppose I'm not fully clear on what you are after as far as extending the WebStorageStateStore class.

pamapa:
> Anyway, I suppose I'm not fully clear on what you are after as far as extending the WebStorageStateStore class.

I was more thinking about this:

your MR:
```ts
public async removeUser(): Promise<void> {
        const logger = this._logger.create("removeUser");
        await this.storeUser(null);
        logger.info("user removed from storage");
        if (this.settings.dpopSettings.enabled) {
            await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");
            logger.debug("removed dpop cyptokeys from storage");
        }
        await this._events.unload();
    }
```

idea:
```ts
public async removeUser(): Promise<void> {
        const logger = this._logger.create("removeUser");
        await this.storeUser(null);
        logger.info("user removed from storage");
        /* not needed, as its part of the new `userStore` and done automatically when calling "await this.storeUser(null);"
        if (this.settings.dpopSettings.enabled) {
            await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");
            logger.debug("removed dpop cyptokeys from storage");
        }*/
        await this._events.unload();
    }
```

...

dbfr3qs:
> > Anyway, I suppose I'm not fully clear on what you are after as far as extending the WebStorageStateStore class.
> 
> I was more thinking about this:
> 
> your MR:
> 
> ```ts
> public async removeUser(): Promise<void> {
>         const logger = this._logger.create("removeUser");
>         await this.storeUser(null);
>         logger.info("user removed from storage");
>         if (this.settings.dpopSettings.enabled) {
>             await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");
>             logger.debug("removed dpop cyptokeys from storage");
>         }
>         await this._events.unload();
>     }
> ```
> 
> idea:
> 
> ```ts
> public async removeUser(): Promise<void> {
>         const logger = this._logger.create("removeUser");
>         await this.storeUser(null);
>         logger.info("user removed from storage");
>         /* not needed, as its part of the new `userStore` and done automatically removed when calling "await this.storeUser(null);"
>         if (this.settings.dpopSettings.enabled) {
>             await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");
>             logger.debug("removed dpop cyptokeys from storage");
>         }*/
>         await this._events.unload();
>     }
> 
> // untested pseudo code only
> export class DPopStorageStateStore extends WebStorageStateStore {
>    public async remove(key: string): Promise<string | null> {
>      const ret = super.remove(key);
>      await IndexedDbCryptoKeyPairStore.remove("oidc.dpop");
>      return ret;
>    }
>    ...
> }
> ```

Okay, yep - and would you like to try and apply the same idea to all the methods on the WebStorateStateStore class - e.g. set, get, getAllKeys such that it can handle DPoP related data as well as he usual user attributes stored in local/session storage?

src/DPoPService.ts
```diff
@@ -0,0 +1,104 @@
...
+ * OAuth2 spec https://datatracker.ietf.org/doc/html/rfc9449.
+ */
+
+export interface GenerateDPoPProofOpts {
+    url: string;
+    accessToken?: string;
+    httpMethod?: string;
+}
+export class DPoPService {
+    public static async generateDPoPProof({
```
pamapa, line 15: please make this a normal service like all thers:
- constructor
- no static methods
- instantiate it within the `UserManager` constructor
- like `this._silentRenewService = new SilentRenewService(this);`

pamapa:
> Okay, yep - and would you like to try and apply the same idea to all the methods on the WebStorateStateStore class - e.g. set, get, getAllKeys such that it can handle DPoP related data as well as the usual user attributes stored in local/session storage?

if possible yes. the idea is to simplify the dpop code paths and the **not** dpop code paths.
Stuff like:
- User.ts (BTW: not used anywhere)
  ```
 public async dpopProof(url: string, httpMethod?: string): Promise<string> {
        return await DPoPService.generateDPoPProof({ url, accessToken: this.access_token, httpMethod: httpMethod } );
    }
  ```
- it contains directly code of `export class IndexedDbCryptoKeyPairStore` or hold the only reference to that object.

src/User.ts
```diff
@@ -124,4 +125,8 @@ export class User {
         Logger.createStatic("User", "fromStorageString");
         return new User(JSON.parse(storageString));
     }
+
+    public async dpopProof(url: string, httpMethod?: string): Promise<string> {
```
pamapa, line 129: not used at all -> remove

dbfr3qs: Ok, but as pointed out on my original PR we need to expose the dpopProof somehow to the consumer so that it can be used as a request header for subsequent authenticated fetch requests to protected resources. Where would you suggest this be done? Here is an [example](https://github.com/dbfr3qs/react-oidc-context/blob/91f84e7118a3f7d5bdda4115536a79f0b858a587/example/index.tsx#L30-L38) using the react-oidc-client.

pamapa: > Ok, but as pointed out on my [original PR](https://github.com/authts/oidc-client-ts/pull/1361#discussion_r1465720125) we need to expose the generation of dpop proofs somehow to the consumer so that a proof can be used in the DPoP request header for subsequent authenticated fetch requests to protected resources. Where would you suggest this be done? Here is an [example](https://github.com/dbfr3qs/react-oidc-context/blob/91f84e7118a3f7d5bdda4115536a79f0b858a587/example/index.tsx#L30-L38) using the react-oidc-client

Ah i forgot that, makes totally sense.

I propose adding it with a describing comment to the `UserManager` class. This way `DPoPService` does not need to be static. The parameter to this function must probably be the user object, to have access to the access_token.

dbfr3qs: Cool - that makes perfect sense 👍

dbfr3qs:
> > Okay, yep - and would you like to try and apply the same idea to all the methods on the WebStorateStateStore class - e.g. set, get, getAllKeys such that it can handle DPoP related data as well as the usual user attributes stored in local/session storage?
> 
> if possible yes. the idea is to simplify the dpop code paths and the **not** dpop code paths. Stuff like:
> 
> * User.ts (BTW: not used anywhere)
> 
> public async dpopProof(url: string, httpMethod?: string): Promise { return await DPoPService.generateDPoPProof({ url, accessToken: this.access_token, httpMethod: httpMethod } ); }
> 
> ```
> - it contains directly code of `export class IndexedDbCryptoKeyPairStore` or hold the only reference to that object.
> ```

So I gave this a go but it's not possible to extend the WebStorageStateStore class with a DPoPStorageStateStore class without changing the types on the WebStorageState class (and the interface) to allow returning a CryptoKeyPair on the get method... which seems a bit gross, e.g.

```
public async get(key: string): Promise<string | null | CryptoKeyPair> {
        if (key === this.dpop_key) {
            return await IndexedDbCryptoKeyPairStore.get(this.dpop_key);
        } else {
            return await super.get(key);
        }
    }
```

doesn't work unless I change the signature of the parent WebStorageStateStore class to also return `<string | null | CryptoKeyPair>`

One option might be to alter the WebStorateStateStore to handle DPoP as well as what it already does (we'd still need to alter the StateStore interface).

What do you think?

pamapa:
> One option might be to alter the WebStorateStateStore to handle DPoP as well as what it already does (we'd still need to alter the StateStore interface).

Maybe by adding a "new/additional" get like `getDpop(): Promise<null | CryptoKeyPair>`

dbfr3qs:
> > One option might be to alter the WebStorateStateStore to handle DPoP as well as what it already does (we'd still need to alter the StateStore interface).
> 
> Maybe by adding a "new/additional" get like `getDpop(): Promise<null | CryptoKeyPair>`

Ok, I have had a rough go at adding a DPoPStorageStateStore for you to look at. I altered the StateStore interface to accept the CryptoKeyPair type, hopefully that makes sense. Also added the additional getDPoP method and inject the instance of the store into the DPoPService now.

Have a look and let me know what you think. If you're happy with the approach, I will move the IndexedDBCryptoKeyPairStore logic directly into the DPoPStorageStateStore and incorporate any other feedback you may have.

I'll be travelling for the next few weeks so may not be able to do too much until May 👍

pamapa:
> Have a look and let me know what you think. If you're happy with the approach, I will move the IndexedDBCryptoKeyPairStore logic directly into the DPoPStorageStateStore and incorporate any other feedback you may have.
> 
> I'll be travelling for the next few weeks so may not be able to do too much until May 👍

I will have a deeper look at this in the next days. Thanks for keeping up on this.

dbfr3qs:
> > Have a look and let me know what you think. If you're happy with the approach, I will move the IndexedDBCryptoKeyPairStore logic directly into the DPoPStorageStateStore and incorporate any other feedback you may have.
> > I'll be travelling for the next few weeks so may not be able to do too much until May 👍
> 
> I will have a deeper look at this in the next days. Thanks for keeping up on this.

Cool - I am back from travelling so can start working on this again when you have had a look 👍

dbfr3qs:
Any thoughts on the latest commit @pamapa ? I am keen to progress this.

pamapa:
Was busy the last weeks, I have checked out your branch and I am now looking at this code and have some thoughts and time in the future to work with you...

General design
- I am afraid of the additional code size for people who do not need DPoP, thus my first idea was to move everything into a sub class of `UserManager`, like `DPoPUserManager`, but when looking at the new code this does not seem feasible. The additional code is tight into the existing code.
- By design `OidcClient` is for the protocol and `UserManager` is for "user related" stuff. DPoP is mostly protocol stuff except the access token from the user.
- Thus maybe it really makes sense to have all but `public async dpopProof(` (access token needed), in the `OidcClient` class like you had in the beginning (sorry about that).

DPoPService:
- i am not so sure it is worth to have this service at all, at the end its a bunch of crypto utils and getting the keys.
- we could split the utils and getting the keys.
- crypto utils code could be added to the `CryptoUtils` and adding a parameter to pass the `keyPair`

DPoPStorageStateStore/IndexedDbCryptoKeyPairStore:
- when i look at the current (from me previously proposed idea) implementation of extending the `WebStorageStateStore`, its really hard to read and seems over complicated...
- at the end is probably simpler if we simply add an additional optional store like `dpopStore`.
- like this we can merge DPoPStorageStateStore & IndexedDbCryptoKeyPairStore into one and keep its as simple as possible
- if somebody needs a different implementation of `dpopStore` he can easily do so.

DPoPSettings:
- lets rename dpopSettings to dpop (it is already within the settings)
- drop enable property (if dpop is undefined its like false if not its enabled)
- add store

Something like:
```ts
let dpopJkt: string | undefined;
if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
  const keys = await this.settings.dpop.store.get();
  dpopJkt = await CryptoUtils.generateDPoPJkt({ keys });
}
```

dbfr3qs:
> Was busy the last weeks, I have checked out your branch and I am now looking at this code and have some thoughts and time in the future to work with you...
> 
> General design
> 
> * I am afraid of the additional code size for people who do not need DPoP, thus my first idea was to move everything into a sub class of `UserManager`, like `DPoPUserManager`, but when looking at the new code this does not seem feasible. The additional code is tight into the existing code.
> * By design `OidcClient` is for the protocol and `UserManager` is for "user related" stuff. DPoP is mostly protocol stuff except the access token from the user.
> * Thus maybe it really makes sense to have all but `public async dpopProof(` (access token needed), in the `OidcClient` class like you had in the beginning (sorry about that).
> 
> DPoPService:
> 
> * i am not so sure it is worth to have this service at all, at the end its a bunch of crypto utils and getting the keys.
> * we could split the utils and getting the keys.
> * crypto utils code could be added to the `CryptoUtils` and adding a parameter to pass the `keyPair`
> 
> DPoPStorageStateStore/IndexedDbCryptoKeyPairStore:
> 
> * when i look at the current (from me previously proposed idea) implementation of extending the `WebStorageStateStore`, its really hard to read and seems over complicated...
> * at the end is probably simpler if we simply add an additional optional store like `dpopStore`.
> * like this we can merge DPoPStorageStateStore & IndexedDbCryptoKeyPairStore into one and keep its as simple as possible
> * if somebody needs a different implementation of `dpopStore` he can easily do so.
> 
> DPoPSettings:
> 
> * lets rename dpopSettings to dpop (it is already within the settings)
> * drop enable property (if dpop is undefined its like false if not its enabled)
> * add store
> 
> Something like:
> 
> ```ts
> let dpopJkt: string | undefined;
> if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
>   const keys = await this.settings.dpop.store.get();
>   dpopJkt = await CryptoUtils.generateDPoPJkt({ keys });
> }
> ```

Thank you for the detailed feedback @pamapa, that is excellent and exactly what I was looking for. At a glance, that all makes sense and I will update this PR accordingly over the next week.

One minor clarifying implementation question for you regarding the new `dpopStore`: does it matter if it implements the storageStage interface? It's minor, but will require extending the interface as I have already done (to handle the `CryptoKeyPair` type).

pamapa:
> One minor clarifying implementation question for you regarding the new `dpopStore`: does it matter if it implements the storageStage interface? It's minor, but will require extending the interface as I have already done (to handle the `CryptoKeyPair` type).

If we create a new storage type we have some flexibility. I guess when we decouple from `StateStore` its getting simpler for existing `string` cases and `CryptoKeyPair`.
My hope is that when someone does not use dpop, that store is removed when doing tree-shaking in production builds, as its not used/referenced directly. Its only passed via settings.

src/UserManagerSettings.ts
```diff
@@ -55,7 +55,6 @@ export interface UserManagerSettings extends OidcClientSettings {
     validateSubOnSilentRenew?: boolean;
     /** Flag to control if id_token is included as id_token_hint in silent renew calls (default: false) */
     includeIdTokenInSilentRenew?: boolean;
-
```
pamapa, line 58: please keep this newline

src/UserManager.ts
```diff
@@ -289,15 +311,22 @@ export class UserManager {
         if (user?.refresh_token) {
             logger.debug("using refresh token");
             const state = new RefreshState(user as Required<User>);
+            const extraHeaders: Record<string, ExtraHeader> = {};
```
pamapa, line 314: extraHeaders is now empty, thus we do not need it (i guess that is a leftover from refactoring...)

src/UserManager.ts
```diff
@@ -170,9 +173,16 @@ export class UserManager {
...
+
+        let dpopJkt: string | undefined;
+        if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
+            dpopJkt = await this.generateDPoPJkt(this.settings.dpop);
+        }
+
         const handle = await this._redirectNavigator.prepare({ redirectMethod });
         await this._signinStart({
             request_type: "si:r",
+            dpopJkt: dpopJkt,
```
pamapa, line 185: we can use a shorthand here: `dpopJkt,`

src/OidcClientSettings.ts
```diff
@@ -271,5 +289,9 @@ export class OidcClientSettingsStore {
         this.extraQueryParams = extraQueryParams;
         this.extraTokenParams = extraTokenParams;
         this.extraHeaders = extraHeaders;
+        this.dpop = dpop;
```
pamapa, line 292: please add a newline between dpop "block" and extra "block"

src/OidcClient.ts
```diff
@@ -171,10 +174,32 @@ export class OidcClient {
 
         const { state, response } = await this.readSigninResponseState(url, true);
         logger.debug("received state from storage; validating response");
+
+        if (this.settings.dpop && this.settings.dpop.store) {
+            const dpopProof = await this.getDpopProof(this.settings.dpop.store);
+            extraHeaders = { ...extraHeaders, "DPoP": dpopProof };
+        }
```
pamapa, line 181: please add a newline after `}`

pamapa:
> @pamapa I believe I have integrated most, if not all, of your most recent suggestions. I have also rebased with main. Please let me know your thoughts 🙏

Looks awesome, its very close to be merged. I have found some little style stuff (see below). Thanks a lot for being patient and taking care!

Question:
- in `OidcClientSettings.ts` i see `this.dpop.store = new IndexedDbDPoPStore();` It is done now like the other stores. The difference is that the other stores are essential, where the dpop is not. When we would **not** create here a new object, for everybody who does not use dpop, the `IndexedDbDPoPStore` could be tree-shaked (can you verify this?). This would mean when you use dpop you simple must pass that store via config, which is very easy... What do you think?
- The above would also make this `if (this.settings.dpop && this.settings.dpop.store) {` into this `if (this.settings.dpop) {` as `store` is no longer an optional property in dpop...

src/UserManager.ts
```diff
@@ -170,9 +173,16 @@ export class UserManager {
             redirectMethod,
             ...requestArgs
         } = args;
+
+        let dpopJkt: string | undefined;
+        if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
```
pamapa, line 178: -> `if (this.settings.dpop?.bind_authorization_code) {`

src/UserManager.ts
```diff
@@ -231,6 +246,12 @@ export class UserManager {
      */
     public async signinPopup(args: SigninPopupArgs = {}): Promise<User> {
         const logger = this._logger.create("signinPopup");
+
+        let dpopJkt: string | undefined;
+        if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
```
pamapa, line 251: -> `if (this.settings.dpop?.bind_authorization_code) {`

src/UserManager.ts
```diff
@@ -289,15 +311,22 @@ export class UserManager {
...
                 redirect_uri: requestArgs.redirect_uri,
                 resource: requestArgs.resource,
                 extraTokenParams: requestArgs.extraTokenParams,
                 timeoutInSeconds: silentRequestTimeoutInSeconds,
+                extraHeaders: extraHeaders,
             });
         }
 
+        let dpopJkt: string | undefined;
+        if (this.settings.dpop && this.settings.dpop.bind_authorization_code) {
```
pamapa, line 326: -> `if (this.settings.dpop?.bind_authorization_code) {`

dbfr3qs:
> > @pamapa I believe I have integrated most, if not all, of your most recent suggestions. I have also rebased with main. Please let me know your thoughts 🙏
> 
> Looks awesome, its very close to be merged. I have found some little style stuff (see above & below). Thanks a lot for being patient and taking care!
> 
> Question:
> 
> * in `OidcClientSettings.ts` i see `this.dpop.store = new IndexedDbDPoPStore();` It is done now like the other stores. The difference is that the other stores are essential, where the dpop is not. When we would **not** create here a new object, for everybody who does not use dpop, the `IndexedDbDPoPStore` could be tree-shaked (can you verify this?). This would mean when you use dpop you simple must pass that store via config, which is very easy... What do you think?
> * The above would also make this `if (this.settings.dpop && this.settings.dpop.store) {` into this `if (this.settings.dpop) {` as `store` is no longer an optional property in dpop...

I see what you are saying: we would just need to be clear that an implementation of the `dpopStore` must be supplied, and that we provide an implementation that makes use of IndexDb as a default. I don't have a problem with that really, and you are right that in its current form the store is not tree shaked and adds about 1kb:
image

As far as overall build size goes, there is a 5kb difference between master and this branch when I bundle it
into a sample application using Parcel.

pamapa:
> EDIT: I can confirm that changing the `dpopStore` to a non optional configuration parameter when dpop is enabled does allow the `IndexedDbDPoPStore` to be tree shaked. Do you think we should add a check that the store is actually configured when `drop` is enabled, and throw an exception if not?

We can explicitly throw an exception if the store is undefined, despite the type would enforce to not be undefined. Maybe hidden in the settings class...

> I've also fixed all the style issues you high lighted. Would like me to add some documentation regarding the feature and how to use it to this PR?

Would be nice if you could add a new section in the docs/index.md. This can also be done in a next MR when this one has been merged...

dbfr3qs:
> > EDIT: I can confirm that changing the `dpopStore` to a non optional configuration parameter when dpop is enabled does allow the `IndexedDbDPoPStore` to be tree shaked. Do you think we should add a check that the store is actually configured when `drop` is enabled, and throw an exception if not?
> 
> We can explicitly throw an exception if the store is undefined, despite the type would enforce to not be undefined. Maybe hidden in the settings class...
> 
> > I've also fixed all the style issues you high lighted. Would like me to add some documentation regarding the feature and how to use it to this PR?
> 
> Would be nice if you could add a new section in the docs/index.md. This can also be done in a next MR when this one has been merged...

No problem, happy to do the docs in a different PR. For now I have annotated the `IndexedDbDPoPStore` as @public with a short description.

I have also updated with main and added a check in `OidcSettings` to throw an exception if `drop` is configured without a store.

src/UserManager.ts
```diff
@@ -153,6 +153,9 @@ export class UserManager {
     public async removeUser(): Promise<void> {
         const logger = this._logger.create("removeUser");
         await this.storeUser(null);
+        if (this.settings.dpop) {
```
pamapa, line 156: We should move that dpop remove into the  `storeUser` (`user === null` case`).

pamapa (approved):

psgfaa:
Thank you for the contribution.  I noticed in your comments you removed the nonce option.  We implement it here within our organization.  Is it possible you could add the nonce back in?  We get passed the DPoP Proof but then are warned we need to supply the nonce value.  

Can you provide me with a workaround if you do not add it to the source?  Any help is greatly appreciated and I do thank you for the contribution.

dbfr3qs:
> Thank you for the contribution. I noticed in your comments you removed the nonce option. We implement it here within our organization. Is it possible you could add the nonce back in? We get past the DPoP Proof but then are warned we need to supply the nonce value.
> 
> Can you provide me some help or a workaround? Any help is greatly appreciated and I do thank you for the contribution.

Hi there, yes I originally added a rough implementation but opted to remove it in favour of adding it in a separate PR (the PR was getting quite large, there was disagreement regarding the implementation and it is optional in the spec).

I will open another PR next week. Perhaps you would be willing to help test and review it?

psgfaa:
> > Thank you for the contribution. I noticed in your comments you removed the nonce option. We implement it here within our organization. Is it possible you could add the nonce back in? We get past the DPoP Proof but then are warned we need to supply the nonce value.
> > Can you provide me some help or a workaround? Any help is greatly appreciated and I do thank you for the contribution.
> 
> Hi there, yes I originally added a rough implementation but opted to remove it in favour of adding it in a separate PR (the PR was getting quite large, there was disagreement regarding the implementation and it is optional in the spec).
> 
> I will open another PR next week. Perhaps you would be willing to help test and review it?

Thank you once again, great work!   I found your oidc-client-context and was able to get it working in my code.  But got stuck on the DPoP nonce requirement.

Yes absolutely I can test it and help anyway I can.  Again, appreciate your work and the author of the library - great work !

dbfr3qs:
@psgfaa please take a look at #1569 🙏

PR #1490, opened 2024-05-02
fix: fix the return type of signinCallback

noshiro-pf:
The `Promise<User | void>` return type of `signinCallback` was not convenient to use, so I changed it to the `Promise<User | undefined>` type. There is no runtime change at all since it is just a cast.

### Checklist


pamapa (approved):

pamapa:
thanks for contributing

PR #1569, opened 2024-07-02
Add dpop nonce support

dbfr3qs:
This is a first cut aimed at adding `DPoP` nonce support. I initially removed this feature in the original [PR](https://github.com/authts/oidc-client-ts/pull/1461) and this is a follow up.

This is very experimental and open to change. @pamapa I would really appreciate your thoughts as to how this could be done more cleanly 🙏🏼

The main functionality is supporting [Authorization Server Provided Nonces](https://datatracker.ietf.org/doc/html/rfc9449#name-authorization-server-provid).

According to the spec, if/when an AS requires the use of a nonce and a token request is made without a nonce claim within the DPoP proof JWT, the AS responds with a specific 400 bad request that includes a response header `dpop-nonce` which contains the nonce value to be used. The client is then expected to regenerate its DPoP proof again with the included nonce, and then retry the token request.

>Upon receiving the nonce, the client is expected to retry its token request using a DPoP proof including the supplied nonce value in the nonce claim of the DPoP proof. An example unencoded JWT payload of such a DPoP proof including a nonce is shown below.

e.g.

```
HTTP/1.1 400 Bad Request
 DPoP-Nonce: eyJ7S_zG.eyJH0-Z.HX4w-7v

 {
  "error": "use_dpop_nonce",
  "error_description":
    "Authorization server requires nonce in DPoP proof"
 }
[Figure 20](https://datatracker.ietf.org/doc/html/rfc9449#figure-20): [HTTP 400 Response to a Token Request without a Nonce](https://datatracker.ietf.org/doc/html/rfc9449#name-http-400-response-to-a-toke)
Other HTTP headers and JSON fields MAY also be included in the error response, but there MUST NOT be more than one DPoP-Nonce header.

Upon receiving the nonce, the client is expected to retry its token request using a DPoP proof including the supplied nonce value in the nonce claim of the DPoP proof. An example unencoded JWT payload of such a DPoP proof including a nonce is shown below.

 {
  "jti": "-BwC3ESc6acc2lTc",
  "htm": "POST",
  "htu": "https://server.example.com/token",
  "iat": 1562262616,
  "nonce": "eyJ7S_zG.eyJH0-Z.HX4w-7v"
 }

```

In order to accomodate this I have added a new abstraction `DPoPState`, which is an object that contains both the CryptoKeyPair and the optional nonce value. I have altered the `DPoPStore` interface to accomodate this as the nonce can be used repeatedly for token requests until the AS decides to invalidate.

Which brings me to my next point, the spec is quite open as to how an AS may [provide a new nonce](https://datatracker.ietf.org/doc/html/rfc9449#name-providing-a-new-nonce-value).

>The authorization server MAY supply the new nonce in the same way that the initial one was supplied: by using a DPoP-Nonce HTTP header in the response. The DPoP-Nonce HTTP header field uses the nonce syntax defined in [Section 8.1](https://datatracker.ietf.org/doc/html/rfc9449#NonceSyntax). Each time this happens, it requires an extra protocol round trip.

>A more efficient manner of supplying a new nonce value is also defined by including a DPoP-Nonce HTTP header in the HTTP 200 (OK) response from the previous request. The client MUST use the new nonce value supplied for the next token request and for all subsequent token requests until the authorization server supplies a new nonce.

I attempted to try and add a way to handle a new nonce upon a 200 response but it was going to be messy and potentially involved altering the `SigninResponse`  class, which didn't feel right. I looked at how IdentityServer issues new nonces and they just [re-use the 400 bad request method](https://github.com/DuendeSoftware/IdentityServer/blob/c1e1508c0749265dae983601e7ed7b88eac8fe8b/src/IdentityServer/Validation/Default/DefaultDPoPProofValidator.cs#L414-L422). I think we should leave this out for the time being.

In regards to supporting [Resource Server provided nonces](https://datatracker.ietf.org/doc/html/rfc9449#name-resource-server-provided-no), that is a lot simpler as the client should manage storing and renewing the resource provided nonce as it will be returned on a client initiated fetch request to the RS - it has nothing to do with `oidc-client-ts`. For now I have opted to provide a way to supply the nonce as an [optional parameter when creating the DPoPProof on the user manager class](https://github.com/authts/oidc-client-ts/compare/main...dbfr3qs:oidc-client-ts:add-dpop-nonce-support?expand=1#diff-be27343df6f02daa157014989cbc33093c0d605f2103d55d569a6dd1ab3fdf84R820). We can revisit if we want to offer a way for clients to store the RS provided nonce values and/or provide helpers to manage them, but this interaction falls outside the relationship between the client and the AS. 

### Checklist

### DIFF
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -15,7 +15,8 @@ import { SigninState } from "./SigninState";
 import { State } from "./State";
 import { TokenClient } from "./TokenClient";
 import { ClaimsService } from "./ClaimsService";
-import type { DPoPStore } from "./DPoPStore";
+import { DPoPState, type DPoPStore } from "./DPoPStore";
+import { ErrorDPoPNonce } from "./errors/ErrorDPoPNonce";
 
 /**
  * @public
@@ -182,24 +183,39 @@ export class OidcClient {
             extraHeaders = { ...extraHeaders, "DPoP": dpopProof };
         }
 
-        await this._validator.validateSigninResponse(response, state, extraHeaders);
+        try {
+            await this._validator.validateSigninResponse(response, state, extraHeaders);
+        }
+        catch (err) {
+            if (err instanceof ErrorDPoPNonce) {
+                const dpopProof = await this.getDpopProof(this.settings.dpop!.store, err.nonce);
+                extraHeaders!["DPoP"] = dpopProof;
+                await this._validator.validateSigninResponse(response, state, extraHeaders);
+            } else {
+                throw err;
+            }
+        }
+
         return response;
     }
 
-    async getDpopProof(dpopStore: DPoPStore): Promise<string> {
+    async getDpopProof(dpopStore: DPoPStore, nonce?: string): Promise<string> {
         let keyPair: CryptoKeyPair;
+        let dpopState: DPoPState;
 
         if (!(await dpopStore.getAllKeys()).includes(this.settings.client_id)) {
             keyPair = await CryptoUtils.generateDPoPKeys();
-            await dpopStore.set(this.settings.client_id, keyPair);
+            dpopState = new DPoPState(keyPair);
+            await dpopStore.set(this.settings.client_id, dpopState);
         } else {
-            keyPair = await dpopStore.get(this.settings.client_id);
+            dpopState = await dpopStore.get(this.settings.client_id);
         }
 
         return await CryptoUtils.generateDPoPProof({
             url: await this.metadataService.getTokenEndpoint(false),
             httpMethod: "POST",
-            keyPair: keyPair,
+            keyPair: dpopState.keys,
+            nonce: nonce,
         });
     }
 
@@ -244,16 +260,36 @@ export class OidcClient {
             extraHeaders = { ...extraHeaders, "DPoP": dpopProof };
         }
 
-        const result = await this._tokenClient.exchangeRefreshToken({
-            refresh_token: state.refresh_token,
-            // provide the (possible filtered) scope list
-            scope,
-            redirect_uri,
-            resource,
-            timeoutInSeconds,
-            extraHeaders,
-            ...extraTokenParams,
-        });
+        let result;
+        try {
+            result = await this._tokenClient.exchangeRefreshToken({
+                refresh_token: state.refresh_token,
+                // provide the (possible filtered) scope list
+                scope,
+                redirect_uri,
+                resource,
+                timeoutInSeconds,
+                extraHeaders,
+                ...extraTokenParams,
+            });
+        } catch (err) {
+            if (err instanceof ErrorDPoPNonce) {
+                extraHeaders!["DPoP"] = await this.getDpopProof(this.settings.dpop!.store, err.nonce);
+                result = await this._tokenClient.exchangeRefreshToken({
+                    refresh_token: state.refresh_token,
+                    // provide the (possible filtered) scope list
+                    scope,
+                    redirect_uri,
+                    resource,
+                    timeoutInSeconds,
+                    extraHeaders,
+                    ...extraTokenParams,
+                });
+            } else {
+                throw err;
+            }
+        }
+
         const response = new SigninResponse(new URLSearchParams());
         Object.assign(response, result);
         logger.debug("validating response", response);
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -205,10 +205,16 @@ export class OidcClient {
 
         if (!(await dpopStore.getAllKeys()).includes(this.settings.client_id)) {
             keyPair = await CryptoUtils.generateDPoPKeys();
-            dpopState = new DPoPState(keyPair);
+            dpopState = new DPoPState(keyPair, nonce);
             await dpopStore.set(this.settings.client_id, dpopState);
         } else {
             dpopState = await dpopStore.get(this.settings.client_id);
+
+            // if the server supplied nonce has changed since the last request, update the nonce
+            if (dpopState.nonce !== nonce && nonce) {
+                dpopState.nonce = nonce;
+                await dpopStore.set(this.settings.client_id, dpopState);
+            }
         }
 
         return await CryptoUtils.generateDPoPProof({
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -221,7 +221,7 @@ export class OidcClient {
             url: await this.metadataService.getTokenEndpoint(false),
             httpMethod: "POST",
             keyPair: dpopState.keys,
-            nonce: nonce,
+            nonce: dpopState.nonce,
         });
     }
 
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -183,12 +183,20 @@ export class OidcClient {
             extraHeaders = { ...extraHeaders, "DPoP": dpopProof };
         }
 
+        /**
+         * The DPoP spec describes a method for Authorization Servers to supply a nonce value
+         * in order to limit the lifetime of a given DPoP proof.
+         * See https://datatracker.ietf.org/doc/html/rfc9449#name-authorization-server-provid
+         * This involves the AS returning a 400 bad request with a DPoP-Nonce header containing
+         * the nonce value. The client must then retry the request with a recomputed DPoP proof
+         * containing the supplied nonce value.
+         */
         try {
             await this._validator.validateSigninResponse(response, state, extraHeaders);
         }
         catch (err) {
-            if (err instanceof ErrorDPoPNonce) {
-                const dpopProof = await this.getDpopProof(this.settings.dpop!.store, err.nonce);
+            if (err instanceof ErrorDPoPNonce && this.settings.dpop) {
+                const dpopProof = await this.getDpopProof(this.settings.dpop.store, err.nonce);
                 extraHeaders!["DPoP"] = dpopProof;
                 await this._validator.validateSigninResponse(response, state, extraHeaders);
             } else {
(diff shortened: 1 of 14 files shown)


psgfaa:
This looks great, I will try to implement and test it today

src/OidcClient.ts
```diff
@@ -244,16 +266,36 @@ export class OidcClient {
...
+                scope,
+                redirect_uri,
+                resource,
+                timeoutInSeconds,
+                extraHeaders,
+                ...extraTokenParams,
+            });
+        } catch (err) {
+            if (err instanceof ErrorDPoPNonce) {
+                extraHeaders!["DPoP"] = await this.getDpopProof(this.settings.dpop!.store, err.nonce);
```
pamapa, line 283: Why not doing this before hand? I dislike doing `exchangeRefreshToken` twice...

src/OidcClient.ts
```diff
@@ -182,24 +183,45 @@ export class OidcClient {
...
         }
 
-        await this._validator.validateSigninResponse(response, state, extraHeaders);
+        try {
+            await this._validator.validateSigninResponse(response, state, extraHeaders);
+        }
+        catch (err) {
+            if (err instanceof ErrorDPoPNonce) {
+                const dpopProof = await this.getDpopProof(this.settings.dpop!.store, err.nonce);
+                extraHeaders!["DPoP"] = dpopProof;
```
pamapa, line 192: Why not doing this before hand? I dislike doing `validateSigninResponse` twice...

dbfr3qs: Apologies, perhaps I wasn't Clea enough in the PR description: that's how it works according to the spec. 

The nonce is supplied by the server, and the mechanism is through a bad request response from the token endpoint when fetching a token. Example from the [spec](https://datatracker.ietf.org/doc/html/rfc9449#name-authorization-server-provid):

```
For example, in response to a token request without a nonce when the authorization server requires one, the authorization server can respond with a DPoP-Nonce value such as the following to provide a nonce value to include in the DPoP proof
HTTP/1.1 400 Bad Request
 DPoP-Nonce: eyJ7S_zG.eyJH0-Z.HX4w-7v

 {
  "error": "use_dpop_nonce",
  "error_description":
    "Authorization server requires nonce in DPoP proof"
 }
[Figure 20](https://datatracker.ietf.org/doc/html/rfc9449#figure-20): [HTTP 400 Response to a Token Request without a Nonce](https://datatracker.ietf.org/doc/html/rfc9449#name-http-400-response-to-a-toke)
Other HTTP headers and JSON fields MAY also be included in the error response, but there MUST NOT be more than one DPoP-Nonce header.

Upon receiving the nonce, the client is expected to retry its token request using a DPoP proof including the supplied nonce value in the nonce claim of the DPoP proof. An example unencoded JWT payload of such a DPoP proof including a nonce is shown below.

 {
  "jti": "-BwC3ESc6acc2lTc",
  "htm": "POST",
  "htu": "https://server.example.com/token",
  "iat": 1562262616,
  "nonce": "eyJ7S_zG.eyJH0-Z.HX4w-7v"
 }
```

If we already have [the nonce stored,](https://github.com/authts/oidc-client-ts/pull/1569/files#diff-c833a002f3fa487666ed42d9742ea5ae2fa4ada00227943616583ed0c3204120R202-R209) we will not make the request twice unless the server responds that the nonce is invalid.

pamapa: Understood. Thanks for clarification.

For future reads of the code it is very helpful to have the right spec section documented near the affected code. Maybe you can add the link you provided above as a comment before the re-try blocks...

dbfr3qs: No problem, that is a fair comment and a good idea 👍

psgfaa:
I apologize if this is the wrong place to put this comment.   Chris I pulled your latest code (14 files changed), rebuilt the npm and put the rebuilt package in my application.  Everything builds fine, however I still get the same message "The DPoP proof JWT header is missing.".

I do get a prompt and I do login, then I get that message.   Do you have an example project that I can test this with?  Thank you.

dbfr3qs:
> I apologize if this is the wrong place to put this comment. Chris I pulled your latest code (14 files changed), rebuilt the npm and put the rebuilt package in my application. Everything builds fine, however I still get the same message "The DPoP proof JWT header is missing.".
> 
> I do get a prompt and I do login, then I get that message. Do you have an example project that I can test this with? Thank you.

Yes, I have been using the `react-oidc-context` example app and the DPoP example from the `IdentityServer` Samples repo to test. You can try using these branches on my forks [here](https://github.com/dbfr3qs/Samples/tree/DPoP-oidc-client-ts/IdentityServer/v6/DPoP) and [here](https://github.com/dbfr3qs/react-oidc-context/tree/DPoP-oidc-client-ts/example). You will need to make sure that the `react-oidc-context` is [referencing](https://github.com/dbfr3qs/react-oidc-context/blob/ec7d2210ab7b92684c10f9d2d67c882e0de8fbd5/package.json#L43) a local build of `oidc-client-ts` (of this branch) in the correct location on your local machine.

What IdP are you using?

psgfaa:
> > I apologize if this is the wrong place to put this comment. Chris I pulled your latest code (14 files changed), rebuilt the npm and put the rebuilt package in my application. Everything builds fine, however I still get the same message "The DPoP proof JWT header is missing.".
> > I do get a prompt and I do login, then I get that message. Do you have an example project that I can test this with? Thank you.
> 
> Yes, I have been using the `react-oidc-context` example app and the DPoP example from the `IdentityServer` Samples repo to test. You can try using these branches on my forks [here](https://github.com/dbfr3qs/Samples/tree/DPoP-oidc-client-ts/IdentityServer/v6/DPoP) and [here](https://github.com/dbfr3qs/react-oidc-context/tree/DPoP-oidc-client-ts/example). You will need to make sure that the `react-oidc-context` is [referencing](https://github.com/dbfr3qs/react-oidc-context/blob/ec7d2210ab7b92684c10f9d2d67c882e0de8fbd5/package.json#L43) a local build of `oidc-client-ts` (of this branch) in the correct location on your local machine.
> 
> What IdP are you using?

Thank you.  I made a little more progress.  I need to review your oidc-client-context project/example further.   I am connecting to a US Government Open ID Connect internal auth server.   I was able to progress past the DPoP Nonce error message.

They turn the dev servers off around 7PM so I can't get the exact error message I was getting but I will post it in the morning.  But it was something along the lines it could not retrieve the publicKey.

pamapa (approved):

pamapa:
Thanks for contributing. As I am not using the dpop feature I really appreciate if you can help supporting this feature within this library. Maybe you have time to help supporting other features as well...

psgfaa:
Here is the error I am receiving.  I traced it back to line 190 from CryptoUtils.ts.

"Could not retrieve dpop keys from storage: Cannot read properties of undefined (reading 'publicKey')"

dbfr3qs:
> Here is the error I am receiving (I am connecting to my site using HTTPS://). I tried in Chrome and Edge and it throws the same error. It does work in Firefox though and auth.IsAuthenticated shows TRUE.
> 
> I traced it back to line 190 from CryptoUtils.ts.
> 
> "Could not retrieve dpop keys from storage: Cannot read properties of undefined (reading 'publicKey')"
> 
> I am doing a auth.signinRedirect(), in UserManager.ts I see line 177 is where we do a generateDPoPJkt call
> 
> Here is my oidcConfig:
> 
> const oidcConfig = { authority: 'https://mydomain.com/oauth2/default', client_id: 'myclientid', redirect_uri: 'https://mydomain/login/', post_logout_redirect_uri: 'https://mydomain/logout', onSigninCallback: onSigninCallback, response_type: 'code', scope: 'openid profile', automaticSilentRenew: true, monitorSession: true, dpop: { bind_authorization_code: true, store: new IndexedDbDPoPStore() }, };

Ok, can you check what is stored in indexedDb in your browser under `oidc.dpop`?

Here is an example of what you might see:

image

My first though is that it's possible that you have an out of date object stored that may need to be cleared. The missing public key indicates that the object [being retrieved ](https://github.com/authts/oidc-client-ts/blob/741d209cf4fa74c76561f18d748d43aff0e68e90/src/UserManager.ts#L835-L840)is not in the shape we expect.

This change introduced a breaking change in that the object that we store for DPoP changed to include storing the nonce.

If the object in indexedDb looks different, delete it and try logging in again.

If this is not the case, would you mind starting an issue and we can work from there? This will allow any findings to be more discoverable for others in the future.

dbfr3qs:
> Thanks for contributing. As I am not using the dpop feature I really appreciate if you can help supporting this feature within this library. Maybe you have time to help supporting other features as well...

Very happy to help. My next PR will document the DPoP feature and I am happy so support it.

psgfaa:
Thank you!  I checked the IndexDB storage option in Chrome and it only shows IndexDB and no database.  But oddly enough I tried the website/app and it works now.   I tried in Edge and it did show the IndexDb>oidc>dpop and a key.  I deleted the oidc database, retried it in Edge and it works correctly.

Not sure what was going on with Chrome regarding the IndexDb database not showing anything but all browsers work now.  Thank you very much for your hard work. 

If you need me to test anything let me know.

dbfr3qs:
> Thank you! I checked the IndexDB storage option in Chrome and it only shows IndexDB and no database. But oddly enough I tried the website/app and it works now. I tried in Edge and it did show the IndexDb>oidc>dpop and a key. I deleted the oidc database, retried it in Edge and it works correctly.
> 
> Not sure what was going on with Chrome regarding the IndexDb database not showing anything but all browsers work now. Thank you very much for your hard work.
> 
> If you need me to test anything let me know.

Interesting, but good news. Very odd.

In your environment, do resource servers (e.g. APIs that you call) also require the dpop nonce? I'd be interested in any experiences regarding that because it is a client concern (e.g. handling receiving, storing and using the nonce value). If you look at the `react-oidc-context` example I linked you to above, you will see a [quick and dirty implementation](https://github.com/dbfr3qs/react-oidc-context/blob/ec7d2210ab7b92684c10f9d2d67c882e0de8fbd5/example/index.tsx#L43C13-L56C14) that stores the nonce for requests to the API in the browser sessionStorage.

psgfaa:
I saw your callApi method., thank you for that example.  

Yes I am working on the backend web api as well.  I will use C# .NET Core Web API for the backend.  I believe it will need to validate the nonce and dpop values with the auth server on all requests.  That is my next task.  When I get that working I can send you over an example or check it in at github.

dbfr3qs:
> I saw your callApi method., thank you for that example.
> 
> Yes I am working on the backend web api as well. I will use C# .NET Core Web API for the backend. I believe it will need to validate the nonce and dpop values with the auth server on all requests. That is my next task. When I get that working I can send you over an example or check it in at github.

No problem. Take a look at the [Duende example [API DPoP code](https://github.com/DuendeSoftware/Samples/tree/main/IdentityServer/v7/DPoP/Api/DPoP). This extends .NET's built in JwtBearer validation. Good luck 👍

psgfaa:
Thank you!  I will take a look.  I appreciate all the help you have been.

PR #1600, opened 2024-07-25
updated docs about logger

klues:
It seems like there is no `Oidc.Log`, I had to use just `Log`.

However I'm still confused about the correct usage. After this:
```
import { Log } from 'oidc-client-ts';
Log.setLogger(console);
Log.setLevel(Log.DEBUG)
```

I see logs things like `[UserManager] getUser: user loaded`, but still not [this log at token renewal](https://github.com/authts/oidc-client-ts/blob/8181652f609242dd14ee1ae4f6b82a9a5116f966/src/SilentRenewService.ts#L51).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -82,18 +82,19 @@ The [User](classes/User.html) type is returned from the [UserManager](classes/Us
 
 
 # Logging
-The oidc-client-ts library supports logging. You can set a logger by assigning `Oidc.Log.logger` to anything that supports a `info`, `warn`, and `error` methods that accept a params array. By default, no logger is configured.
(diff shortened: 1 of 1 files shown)


pamapa (approved):

pamapa:
> I see logs things like `[UserManager] getUser: user loaded`, but still not [this log at token renewal](https://github.com/authts/oidc-client-ts/blob/8181652f609242dd14ee1ae4f6b82a9a5116f966/src/SilentRenewService.ts#L51).

depends how you hook the callback. Ensure when using another instance that Log is also setup correctly...

pamapa:
thanks for contributing

PR #1691, opened 2024-10-08
Add `removeState` as an option to `processSigninResponse`

ZephireNZ:
Use case:

We have a custom extension to OAuth/OIDC where the `code` is an OTP that the user is sent when the OAuth handshake starts. This code is used instead of the typical randomly generated code in exchange for an access token.

If this code is incorrect, the user cannot retry as the state (including PKCE challenge) gets removed from the store as soon as the code is used. With this change, we can pass `removeState = true` which means even if the validation fails, it can be reattempted with a different code.

Without this change, we have to mimic the behaviour of `processSigninResponse` by:
- Calling `readSigninResponseState` directly with `removeState = false`
- Editing the prototype of `OidcClient` so that `_validator.validateSigninResponse()` can be called by the application.
```ts
OidcClient.prototype.getValidator = function () {
  return this._validator;
};
```

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -346,7 +346,7 @@ export class OidcClient {
     // (undocumented)
     processResourceOwnerPasswordCredentials({ username, password, skipUserInfo, extraTokenParams, }: ProcessResourceOwnerPasswordCredentialsArgs): Promise<SigninResponse>;
     // (undocumented)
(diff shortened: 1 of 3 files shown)


pamapa (approved):

pamapa:
This library can also be used as a SDK, like you do. Thanks for contributing.

Issue #1438, opened 2024-03-14
client_secret required with PKCE


fpue:
I am currently switching to the Authorization Code Grant with PKCE, but when I try calling the signinCallback() method after the redirect, I get the following error: "Error: A client_secret is required.". Which is weird since PKCE should work without secret.
How can I fix this?

Here's my current config:

```typescript
const settings: UserManagerSettings = {
      authority: 'iamthe.authority',
      client_authentication: 'client_secret_basic',
      client_id: 'clientid_would_be_here',
      loadUserInfo: true,
      redirect_uri: `${window.location.origin}/signin-callback`,
      silent_redirect_uri: `${window.location.origin}/signin-callback`,
      post_logout_redirect_uri: `${window.location.origin}/logged-out`,
      response_type: 'code',
      scope: 'openid',
      metadata: {...}
    };
```

Tested on versions: 3.0.0-rc0 and 3.0.1

Happy to share any information that may be useful to fix this.

pamapa:
Most probably a configuration problem on your IdP side and not on client side...

fpue:
I digged further into the code and I think it's these lines of code. Is there a reason for basic requiring a secret, while post doesn't?

https://github.com/authts/oidc-client-ts/blob/14168c006c3073be00580d8253f882dad36658f9/src/TokenClient.ts#L109C12-L115C23

pamapa:
> I digged further into the code and I think it's these lines of code. Is there a reason for basic requiring a secret, while post doesn't?
> 
> https://github.com/authts/oidc-client-ts/blob/14168c006c3073be00580d8253f882dad36658f9/src/TokenClient.ts#L109C12-L115C23

That code path is only taken if `settings.client_authentication` is set. Which you do, I oversight it the first time sorry. Do not set this for code flow with PKCE...

fpue:
But my IdP only supports basic. Switching IdP is currently not an option, so this ticket now is more a feature request than a question.

Please add the ability to use `client_authentication: 'client_secret_basic'` and PKCE at the same time.

pamapa:
> But my IdP only supports basic.

Then you need a secret client side...

PR #1738, opened 2024-11-19
allow empty string for client_secret in TokenClient

### LINKED ISSUE #1068: Support for Basic Auth Client Authentication with client_id Only
I'm currently using the oidc-client-ts library for implementing OAuth 2.0 authorization code flow with PKCE in a React frontend application. The VMWare CSP identity provider I'm working with requires the client_id to be included in the Basic Auth header during the access token request, even if there's no client_secret provided. More specifically, the IDP requires a Base64 string of <client_id>:<empty_value> to be passed in as Basic Auth header during the request

However, the oidc-client-ts library currently supports client authentication using the Basic Auth header with both the client_id and client_secret. It does not provide an option to include the client_id only in the Basic Auth header without a client_secret.

I would like to request support for including the client_id in the Basic Auth header during the access token request, even when no client_secret is provided. This would ensure compliance with the requirements of the VMWare CSP identity provider and other similar implementations.

I also want to seek clarification on whether what this IDP requires is spec compliant, so that I may get them to change their procedures instead.

If there's an easy work-around to this, I would love to know it as well, as I'm currently forced to fork the library in order to get it to work.

Thank you for your support.

ntamotsu:
Closes/fixes #1068 and #1438

### Checklist

This pull request includes changes to the TokenClient class in the src/TokenClient.ts file to improve the handling of the client_secret parameter. The changes ensure that the client_secret is explicitly checked for undefined or null values, rather than using a falsy check. This modification allows empty strings as valid client_secret values when using client_secret_basic authentication method, which was previously rejected by the falsy check.

### DIFF
--- a/src/TokenClient.ts
+++ b/src/TokenClient.ts
@@ -112,7 +112,7 @@ export class TokenClient {
         let basicAuth: string | undefined;
         switch (this._settings.client_authentication) {
             case "client_secret_basic":
-                if (!client_secret) {
+                if (client_secret === undefined || client_secret === null) {
                     logger.throw(new Error("A client_secret is required"));
                     throw null; // https://github.com/microsoft/TypeScript/issues/46972
                 }
@@ -173,7 +173,7 @@ export class TokenClient {
         let basicAuth: string | undefined;
         switch (this._settings.client_authentication) {
             case "client_secret_basic":
-                if (!client_secret) {
+                if (client_secret === undefined || client_secret === null) {
                     logger.throw(new Error("A client_secret is required"));
                     throw null; // https://github.com/microsoft/TypeScript/issues/46972
                 }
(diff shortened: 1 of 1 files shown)


src/TokenClient.ts
```diff
@@ -112,7 +112,7 @@ export class TokenClient {
         let basicAuth: string | undefined;
         switch (this._settings.client_authentication) {
             case "client_secret_basic":
-                if (!client_secret) {
+                if (client_secret === undefined || client_secret === null) {
```
pamapa, line 115: `!client_secret` is the same as `client_secret === undefined || client_secret === null`, both evaluate to true when  `client_secret` is null or undefined

pamapa (approved):

Issue #1743, opened 2024-11-21
Signin with PopUp when inside an IFrame doesnt work.


thejurassic:
I’m running my app within an IFrame, as it’s embedded in the Microsoft Dynamics sidebar. I'm using SignInPopup for authentication, but I’m encountering issues due to how the token is returned to the calling window.

Currently, the popup window returns the token to the calling IFrame via storage, which is blocked by browser security settings in IFrame contexts. For this setup, the token should ideally be sent back to the calling IFrame using the browser's postMessage API instead of relying on storage.

Is there an existing configuration or workaround for this? If not, could support for window messaging be considered in a future update? Given the comprehensive nature of this library, I’m hoping I might be overlooking a solution here.

Issue #1202, opened 2023-10-11
Workaround for Storage Partitioning


szahn:
There is a new feature in browsers called [Storage Partitioning](https://developer.chrome.com/en/docs/privacy-sandbox/storage-partitioning/) which prevents iframes in pages that rely on a login popup to authenticate. Logging into an identity provider via a popup window from an iframe within a host window causes browser storage partitioning to break the authentication flow. For example, when developing a Microsoft Teams Tab single page application, and logging into a third party identity provider with OAuth 2.0 Authorization Code PKCE flow, breaks the process. Disabling this feature via `chrome://flags/#third-party-storage-partitioning` seems to resolve it in Chrome but not in Chromium or other browsers. This issue seems to affect other applications, including [PowerBI embedding](https://github.com/GoogleChromeLabs/privacy-sandbox-dev-support/issues/163). There are some workarounds such as using url parameters or `postMessage` mentioned [in this article](https://textslashplain.com/2023/04/12/auth-flows-in-a-partitioned-world/).

miketaylr:
> There is a new feature in browsers called [Storage Partitioning](https://developer.chrome.com/en/docs/privacy-sandbox/storage-partitioning/) which prevents iframes in pages that rely on a login popup to authenticate

Note that this behavior has shipped in Safari and Firefox for quite some time - Chrome is just catching up. How does this library work in those browsers?

szahn:
We found a temporary workaround by using `postMessage` to send the oidc session from the storage in the popup back to the iframe in the host window. However, it would be ideal if the library would have built-in support for partitioned storage.

pamapa:
@szahn Can you post the workaround here? And it would be nice if you can provide support for this via a merge request.

deanmaster:
hi @szahn it would be great if you can post workaround here and perfectly as @pamapa  said if you provide with merge request

PR #1744, opened 2024-11-22
Make signinPopup work when calling window is iframe

thejurassic:
Closes/fixes #issue

### Checklist

When using signinPopup and your app is within an iframe, the popup cannot access the local storage of the calling window due to  fairly recent changes with partitioning local storage. As such the popup window cant access the stored state in storage from the main calling window.

What my changes do, is if the popup window cant find the SigninState in storage , then it returns undefined for State.
so readSigninResponseState in OidcClient.ts now returns Promise<{ state: SigninState|undefined; response: SigninResponse }
This prevents the Popup window from throwing an error that it couldnt find the state. And it returns the URL which includes the state and the calling window can verify the state instead of doing it in the popup window.

Then in UserManager.ts signinCallback function:
if state returned from await this._client.readSigninResponseState(url); is undefined, we assume we should call signinPopupCallback which sends the url back to the calling window, which itself then verifies state from the url.

I have confirmed that this now works when the calling window is in an iframe, when it previously failed at stage readSigninResponseState.  I have also confirmed that auth fails if the response provides a bad state.

To summarize, the popup window was trying to verify the state in the response from the state in storage, and then when the calling window receives the URL it also verifies the state from storage. Now, if the popup cant get the state from storage it passes the url back and the main window only verifies the url and state from storage.

This allows the library to work when the popup window doesnt have access to the same partitioned local storage as the main calling window when the main calling window is in an iframe.

### DIFF
--- a/src/OidcClient.ts
+++ b/src/OidcClient.ts
@@ -152,7 +152,7 @@ export class OidcClient {
         return signinRequest;
     }
 
-    public async readSigninResponseState(url: string, removeState = false): Promise<{ state: SigninState; response: SigninResponse }> {
+    public async readSigninResponseState(url: string, removeState = false): Promise<{ state: SigninState|undefined; response: SigninResponse }> {
         const logger = this._logger.create("readSigninResponseState");
 
         const response = new SigninResponse(UrlUtils.readParams(url, this.settings.response_mode));
@@ -163,12 +163,8 @@ export class OidcClient {
         }
 
         const storedStateString = await this.settings.stateStore[removeState ? "remove" : "get"](response.state);
-        if (!storedStateString) {
-            logger.throw(new Error("No matching state found in storage"));
-            throw null; // https://github.com/microsoft/TypeScript/issues/46972
-        }
 
-        const state = await SigninState.fromStorageString(storedStateString);
+        const state = storedStateString ? await SigninState.fromStorageString(storedStateString) : undefined;
         return { state, response };
     }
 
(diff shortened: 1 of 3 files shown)


thejurassic:
See issues:
#1743
#1202

src/OidcClient.ts
```diff
@@ -152,7 +152,7 @@ export class OidcClient {
         return signinRequest;
     }
 
-    public async readSigninResponseState(url: string, removeState = false): Promise<{ state: SigninState; response: SigninResponse }> {
+    public async readSigninResponseState(url: string, removeState = false): Promise<{ state: SigninState|undefined; response: SigninResponse }> {
```
pamapa, line 155: please add spaces in-between: `SigninState | undefined`

src/OidcClient.ts
```diff
@@ -183,6 +179,11 @@ export class OidcClient {
             extraHeaders = { ...extraHeaders, "DPoP": dpopProof };
         }
 
+        if (!state) {
+            logger.throw(new Error("No state was found in storage or response"));
+            throw null; //
```
pamapa, line 184: please re-add comment from above `// https://github.com/microsoft/TypeScript/issues/46972`

pamapa:
Thanks for taking care of this feature!

src/UserManager.ts
```diff
@@ -394,6 +394,13 @@ export class UserManager {
      */
     public async signinCallback(url = window.location.href): Promise<User | undefined> {
         const { state } = await this._client.readSigninResponseState(url);
+
+        // if no state from storage, assume signin popup
+        if (state === undefined) {
```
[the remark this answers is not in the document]
pamapa: How are we sure it was a "signin popup request" here? What about "No matching state found in storage" in the non "signin popup" request case?

pamapa:
I guess using `sessionStorage` would make it still work without this changes...

Issue #712, opened 2022-10-07
addUserLoaded not fired on UserManager.getUser() method


waldemarennsaed:
## Issue

Apperently, the `addUserLoaded` event is not being fired whenever the `getUser()` method on the `UserManager` is being called.

## Solution

The `addUserLoaded` event should be loaded on every successful `getUser()` method response.

### General structure

Using the following setup:

* Vue 3
* oidc-client-ts (this project)
* Pinia
* Vue-Router 4
* (TypeScript)

### The implementation

Basically, using the following code-structure:

* `/src`
  * `UserManager.ts`
  * `/store`
    * `auth.ts`
    
**oidc.ts**
```typescript
// all imports
const settings: UserManagerSettings = { // ... }

const userManager: UserManager = new UserManager(settings)

// THIS only executes on signInCallback, never on the getUser() method that I use in `/src/store/auth.ts`
userManager.events.addUserLoaded((user: User) => {
  // retrieve some initial API data, or do a bunch of other stuff
})

export { userManager }
```

**auth.ts**
```typescript
import { pinia } from '@/src/store/index.ts'
import { defineStore } from 'pinia'
import { userManager } from '@/src/oidc.ts'

const authStore = defineStore('auth', {
  actions: {
    async checkIfUserIsAuthed () {
      // first check if your current route is any callback-route
      
      // now check if the user is already authed, by checking for a valid getUser() response !== null
      const user: User|null = await userManager.getUser()
    }
  }
})
```

The `checkIfUserIsAuthed` method get's executed by a route-handler, on every route-visit by using a `beforeRouteChange` handler on `vue-router`.

**The issue:** The `getUser()` method in `/src/store/auth.ts` get's executed on every page-reload after once authenticated, but it never triggers the `userLoaded` event anymore. Only after signing in.

## Finally

Am I doing anything wrong? Is there any misconception I have?

My current solution is using a `Proxy` `apply`-trap on `UserManager.getUser()` which checks if the result is `!== null` and then executes my logic. So I am intercepting/extending the `getUser()` functionality - but I guess the event should still be triggered.

Would be happy for an explanation, did I catch an issue here?

pamapa:
The event `userLoaded` is not fired by intention/design of not breaking the behavior of the predecessor library `oidc-client`. It might could produce an endless loop: When you call `getUser` somewhere in your `userLoaded` callback function.

[...]

PR #1817, opened 2025-01-21
Adds raiseEvent parameter to getUser() method

waldemarennsaed:
## Description

Relates to #712 and provides a solution that is backwards-compatible and will fix an issue for some users.

Adding a new parameter `raiseEvent = false` to the `UserManager.getUser()` method to allow execute the `addUserLoaded()` event handlers after raising the `userLoaded` event.

**Background**: The `userLoaded` event does currently not get raised whenever the `user` value is being read from the local storage, which happens whenever no user data has to be fetched from the authority but from the store. In #712 developers complained about this behavior since some business-logic depends on the `getUser()`/`userLoaded` method/event for .e.g. loading additional user-related data from third-party sources like APIs. In the case that a user is already authenticated and e.g. reloads the page, the storage provides the `getUser()` data but does not raise the `userLoaded` event anymore. That will break the described business-logic and must be avoided with a workaround, e.g. using `Proxy()` which is not preferred as the optimal solution (see https://github.com/authts/oidc-client-ts/issues/712#issuecomment-2601945064 for an implementation example with `Proxy()`).

This solution allows raising the event on purpose whenever the parameter is set to handle further business logic after checking for a users authentication.

## Tests

Added unit-tests for the new parameter which is by default set to `false`.

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -1005,7 +1005,7 @@ export class UserManager {
     protected readonly _events: UserManagerEvents;
     // (undocumented)
     generateDPoPJkt(dpopSettings: DPoPSettings): Promise<string | undefined>;
-    getUser(): Promise<User | null>;
+    getUser(raiseEvent?: boolean): Promise<User | null>;
     // (undocumented)
     protected readonly _iframeNavigator: INavigator;
     // (undocumented)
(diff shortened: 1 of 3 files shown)


waldemarennsaed:
@pamapa any thoughts on this?

waldemarennsaed:
@pamapa I am sorry but I must ask and generate some traffic on this one 🤓

pamapa (approved):

PR #1818, opened 2025-01-22
Add url_state to SignoutRequest

richjyoung:
Extends the fix added for #293 in PR https://github.com/authts/oidc-client-ts/pull/1212 to also cover sign out.

For background I believe the same principles regarding `state` for sign in within the linked issue also apply in the sign out case as well. I am working with a multi-tenant system that has a proxy to resolve the actual hostname for the redirect_uri, which also needs to send a signed-out user back to the application start page on the right hostname.

Without this fix, I would need to configure every potential hostname as a valid redirect URI in the IDP client config (whilst I think Keycloak can allow wildcards, it's not recommended and others such as Entra ID do not).

This change replicates the concatenated internal opaque state value with a custom `url_state` value and delimiter as used in sign in. Tests have been added to ensure the `url_state` can still be added even if the sign out request does not need to persist any client state.

Apologies for not raising an issue first, in tweaking the code to see if it was viable I basically had the PR already written.

### Checklist

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -855,7 +855,7 @@ export type SignoutRedirectArgs = RedirectParams & ExtraSignoutRequestArgs;
 
 // @public (undocumented)
 export class SignoutRequest {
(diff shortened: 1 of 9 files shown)


pamapa (approved):

pamapa:
thanks for contributing

PR #1924, opened 2025-04-03
fix: use BroadcastChannel if no window.opener

### LINKED ISSUE #1896: Popup Login Fails with "No window.opener" Error on Callback
**Problem Statement**

I'm using oidc-client-ts to implement OAuth2 PKCE authentication with a popup login flow. However, when the authentication provider redirects back to my popup callback page, I get the following error (see [source code][1]):

```
No window.opener. Can't complete notification.
```

**What I'm Doing**

I initiate the login flow using signinPopup() in my main app:

```js
  authority: "https://your-oidc-provider.com",
  client_id: "your-client-id",
  redirect_uri: window.location.origin + "/callback",
});

await userManager.signinPopup();
```

I handle the popup callback on a separate page (`/callback`):

```js
  authority: "https://your-oidc-provider.com",
  client_id: "your-client-id",
  redirect_uri: window.location.origin + "/callback",
});

await userManager.signinCallback();
```

**Notes**

- `window.opener  = null` inside the popup console meaning the popup doesn't recognize the main page
- `window.crossOriginIsolated = false` (https://github.com/authts/oidc-client-ts/issues/31#issuecomment-896749759 suggests the popup doesn't work if it's `true`)
- Note I'm integrating with many different providers so creating a singleton UserManager instance isn't feasible
- I see the following error in my logs: "Popup closed by user" (looks it comes from [here](https://github.com/authts/oidc-client-ts/blob/8c836ff9fdd9bbf6a7f72890e7f85d46574e1654/src/navigators/PopupWindow.ts#L62)). But my popup is still open. Not sure if that's related.

What am I doing wrong?

  [1]: https://github.com/authts/oidc-client-ts/blob/8c836ff9fdd9bbf6a7f72890e7f85d46574e1654/src/navigators/PopupWindow.ts#L81-L83

jcphill:
Allows Cross-Origin-Opener-Policy="same-origin", which is what Amazon Cognito returns.

Closes/fixes #1896

### Checklist

- [ X] I have included links for closing relevant issue numbers

### DIFF
--- a/src/navigators/AbstractChildWindow.ts
+++ b/src/navigators/AbstractChildWindow.ts
@@ -61,6 +61,9 @@ export abstract class AbstractChildWindow implements IWindow {
             };
             window.addEventListener("message", listener, false);
             this._disposeHandlers.add(() => window.removeEventListener("message", listener, false));
+            const channel = new BroadcastChannel(`oidc-client-popup-${params.state}`);
+            channel.addEventListener("message", listener, false);
+            this._disposeHandlers.add(() => channel.close());
             this._disposeHandlers.add(this._abort.addHandler((reason) => {
                 this._dispose();
                 reject(reason);
@@ -87,7 +90,7 @@ export abstract class AbstractChildWindow implements IWindow {
         this._disposeHandlers.clear();
     }
 
-    protected static _notifyParent(parent: Window, url: string, keepOpen = false, targetOrigin = window.location.origin): void {
+    protected static _notifyParent(parent: Window | BroadcastChannel, url: string, keepOpen = false, targetOrigin = window.location.origin): void {
         parent.postMessage({
             source: messageSource,
             url,
--- a/src/navigators/AbstractChildWindow.ts
+++ b/src/navigators/AbstractChildWindow.ts
@@ -90,11 +90,24 @@ export abstract class AbstractChildWindow implements IWindow {
         this._disposeHandlers.clear();
     }
 
-    protected static _notifyParent(parent: Window | BroadcastChannel, url: string, keepOpen = false, targetOrigin = window.location.origin): void {
-        parent.postMessage({
+    protected static _notifyParent(parent: Window | null, url: string, keepOpen = false, targetOrigin = window.location.origin): void {
+        const msgData: MessageData = {
             source: messageSource,
             url,
             keepOpen,
-        } as MessageData, targetOrigin);
+        };
+        if (parent) {
+            parent.postMessage(msgData, targetOrigin);
+        } else {
+            const logger = new Logger("_notifyParent");
+            logger.debug("No parent. Using BroadcastChannel.");
+            const state = new URL(url).searchParams.get("state");
+            if (!state) {
+                throw new Error("No parent and no state in URL. Can't complete notification.");
+            }
+            const channel = new BroadcastChannel(`oidc-client-popup-${state}`);
+            channel.postMessage(msgData);
+            channel.close();
+        }
     }
 }
--- a/src/navigators/AbstractChildWindow.ts
+++ b/src/navigators/AbstractChildWindow.ts
@@ -96,10 +96,11 @@ export abstract class AbstractChildWindow implements IWindow {
             url,
             keepOpen,
         };
+        const logger = new Logger("_notifyParent");
         if (parent) {
+            logger.debug("With parent. Using parent.postMessage.");
             parent.postMessage(msgData, targetOrigin);
         } else {
-            const logger = new Logger("_notifyParent");
             logger.debug("No parent. Using BroadcastChannel.");
             const state = new URL(url).searchParams.get("state");
             if (!state) {
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -57,12 +57,15 @@ export class PopupWindow extends AbstractChildWindow {
     public async navigate(params: NavigateParams): Promise<NavigateResponse> {
         this._window?.focus();
 
-        const popupClosedInterval = setInterval(() => {
+        const popupClosedInterval = { id: setInterval(() => {
             if (!this._window || this._window.closed) {
                 this._logger.debug("Popup closed by user or isolated by redirect");
+                clearInterval(popupClosedInterval.id);
+                // appease TypeScript while not clearing twice
+                popupClosedInterval.id = setInterval(() => {}, 1000000);
             }
-        }, checkForPopupClosedInterval);
-        this._disposeHandlers.add(() => clearInterval(popupClosedInterval));
+        }, checkForPopupClosedInterval) };
+        this._disposeHandlers.add(() => clearInterval(popupClosedInterval.id));
 
         return await super.navigate(params);
     }
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -57,15 +57,15 @@ export class PopupWindow extends AbstractChildWindow {
     public async navigate(params: NavigateParams): Promise<NavigateResponse> {
         this._window?.focus();
 
-        const popupClosedInterval = { id: setInterval(() => {
+        const popupClosedInterval = setInterval(() => {
             if (!this._window || this._window.closed) {
                 this._logger.debug("Popup closed by user or isolated by redirect");
-                clearInterval(popupClosedInterval.id);
-                // appease TypeScript while not clearing twice
-                popupClosedInterval.id = setInterval(() => {}, 1000000);
+                clearPopupClosedInterval();
+                this._disposeHandlers.delete(clearPopupClosedInterval);
             }
-        }, checkForPopupClosedInterval) };
-        this._disposeHandlers.add(() => clearInterval(popupClosedInterval.id));
+        }, checkForPopupClosedInterval);
+        const clearPopupClosedInterval = () => clearInterval(popupClosedInterval);
+        this._disposeHandlers.add(clearPopupClosedInterval);
 
         return await super.navigate(params);
     }
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -81,19 +81,9 @@ export class PopupWindow extends AbstractChildWindow {
     }
 
     public static notifyOpener(url: string, keepOpen: boolean): void {
-        if (!window.opener) {
-            const logger = new Logger("PopupWindow");
-            logger.debug("No window.opener. Using BroadcastChannel.");
-            const state = new URL(url).searchParams.get("state");
-            if (!state) {
-                throw new Error("No window.opener and no state in URL. Can't complete notification.");
-            }
-            const channel = new BroadcastChannel(`oidc-client-popup-${state}`);
-            super._notifyParent(channel, url, keepOpen);
-            channel.close();
-            window.close();
-            return;
-        }
         super._notifyParent(window.opener, url, keepOpen);
+        if (!keepOpen && !window.opener) {
+            window.close();
+        }
     }
 }
--- a/src/navigators/PopupWindow.ts
+++ b/src/navigators/PopupWindow.ts
@@ -59,7 +59,7 @@ export class PopupWindow extends AbstractChildWindow {
 
         const popupClosedInterval = setInterval(() => {
             if (!this._window || this._window.closed) {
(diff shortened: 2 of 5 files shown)


edthedev:
@jcphill I sent you some thoughts over another channel. 

I want to add here for consideration - one difference for security outcomes is that CORS header approach is opt-in, while this change affects all library users. There might be value in looking for a custom header before applying this work-around, so that this fail-over method has to be explicitly enabled.

edthedev:
Per further discussion, my idea for an extra header doesn't work for this use case, since we do not control headers sent by Cognito. Sorry I missed that nuance. Thanks @jcphill.

jcphill:
The state string is now included in the channel name to ensure that the channel is unique for each popup and only the opener and the popup can connect to it.

jcphill:
@pamapa if you could approve the workflow I will be sure this is ready to merge when you have time to look at it.

pamapa:
@jcphill thanks for working on this i will soon have time to review this

src/navigators/PopupWindow.ts
```diff
@@ -59,7 +59,7 @@ export class PopupWindow extends AbstractChildWindow {
 
         const popupClosedInterval = setInterval(() => {
             if (!this._window || this._window.closed) {
-                void this._abort.raise(new Error("Popup closed by user"));
```
pamapa, line 62: for the use-case where we have a window this raise makes still sense. Why not change the condition to?
`if (this._window && this._window.closed) {...`

src/navigators/AbstractChildWindow.ts
```diff
@@ -61,6 +61,9 @@ export abstract class AbstractChildWindow implements IWindow {
             };
             window.addEventListener("message", listener, false);
             this._disposeHandlers.add(() => window.removeEventListener("message", listener, false));
+            const channel = new BroadcastChannel(`oidc-client-popup-${params.state}`);
```
pamapa, line 64: `AbstractChildWindow` is also used for `IFrameWindow`, i guess there we do not need the `BroadcastChannel`...

pamapa:
test fails with:
```
PopupWindow › should show error when closePopupWindowAfterInSeconds is greater than 0 and window is not open

    expect(received).rejects.toThrow(expected)

    Expected substring: "Popup blocked by user"
    Received message:   "BroadcastChannel is not defined"

          62 |             window.addEventListener("message", listener, false);
          63 |             this._disposeHandlers.add(() => window.removeEventListener("message", listener, false));
        > 64 |             const channel = new BroadcastChannel(`oidc-client-popup-${params.state}`);
             |                                 ^
          65 |             channel.addEventListener("message", listener, false);
          66 |             this._disposeHandlers.add(() => channel.close());
          67 |             this._disposeHandlers.add(this._abort.addHandler((reason) => {

      at BroadcastChannel (src/navigators/AbstractChildWindow.ts:64:33)
      at PopupWindow.navigate (src/navigators/AbstractChildWindow.ts:37:41)
      at PopupWindow.navigate (src/navigators/PopupWindow.ts:67:28)
      at Object.<anonymous> (src/navigators/PopupWindow.test.ts:221:37)
      at Object.toThrow (node_modules/expect/build/index.js:218:22)
      at Object.<anonymous> (src/navigators/PopupWindow.test.ts:225:39)

FAIL src/navigators/IFrameWindow.test.ts
  ● Test suite failed to run

    Jest worker encountered 4 child process exceptions, exceeding retry limit

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:181:21)
```

src/navigators/PopupWindow.ts
jcphill: Both cases have this._window non-null unless the window.open() call in the constructor returned null (likely because of a popup blocker). It is not possible to distinguish if this._window.closed is true because the popup was closed by the user or because the url passed to super.navigate() below has Cross-Origin-Opener-Policy: same-origin.

I do see that this will repeatedly log the message until the dispose handler is called.

It also looks like the closePopupWindowAfterInSeconds timeout is going to throw an error if this._window.closed is true, when really it is the return value of window.open() that should be checked.

src/navigators/AbstractChildWindow.ts
jcphill: According to https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated an iframe can be similarly isolated by the Cross-Origin-Embedder-Policy header so the BroadcastChannel could be needed in that case as well. If so then most of the new notifyOpener() logic in PopupWindow (except for window.close()) should be moved up to _notifyParent() in AbstractChildWindow. That seems easier and more future-proof than moving this code down into PopupWindow.

pamapa: > If so then most of the new notifyOpener() logic in PopupWindow (except for window.close()) should be moved up to _notifyParent() in AbstractChildWindow. That seems easier and more future-proof than moving this code down into PopupWindow.

Makes sense!

pamapa:
test fails
```
FAIL src/navigators/PopupWindow.test.ts (5.249 s)
  ● PopupWindow › should reject when the window is closed by user

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      121 |     });
      122 |
    > 123 |     it("should reject when the window is closed by user", async () => {
          |     ^
      124 |         const popupWindow = new PopupWindow({});
      125 |
      126 |         const promise = popupWindow.navigate({ url: "http://sts/authorize?x=y", state: "someid" });

      at src/navigators/PopupWindow.test.ts:123:5
      at Object.<anonymous> (src/navigators/PopupWindow.test.ts:18:1)
```

src/navigators/PopupWindow.ts
jcphill: I fixed the repeated log message.

src/navigators/AbstractChildWindow.ts
jcphill: Done.

pamapa:
> @pamapa Ready for review. I removed the failing test and everything else passes. Not rejecting if the popup seems to have closed may be an API change but reflects an underlying browser change so hard to avoid.

The code look good, i will first test this with a real IdP on my side. This is definitely not a bug-fix, i will bump the minor for this. Before doing so i plan a bug-fix release...

src/navigators/PopupWindow.ts
```diff
@@ -57,12 +57,15 @@ export class PopupWindow extends AbstractChildWindow {
     public async navigate(params: NavigateParams): Promise<NavigateResponse> {
         this._window?.focus();
 
-        const popupClosedInterval = setInterval(() => {
+        const popupClosedInterval = { id: setInterval(() => {
```
pamapa, line 60: Within this library we have multiple `setInterval` calls, and never as `{ id: setInterval...}` needed/used, please avoid this here as well

src/navigators/PopupWindow.ts
```diff
@@ -57,12 +57,15 @@ export class PopupWindow extends AbstractChildWindow {
...
             if (!this._window || this._window.closed) {
-                void this._abort.raise(new Error("Popup closed by user"));
+                this._logger.debug("Popup closed by user or isolated by redirect");
+                clearInterval(popupClosedInterval.id);
+                // appease TypeScript while not clearing twice
```
pamapa, line 64: why is this needed? Here TypeScript has no problem with this simpler code:
```ts
...
        const popupClosedInterval = setInterval(() => {
            if (!this._window || this._window.closed) {
                this._logger.debug("Popup closed by user or isolated by redirect");
                clearInterval(popupClosedInterval);
            }
        }, checkForPopupClosedInterval);
        this._disposeHandlers.add(() => clearInterval(popupClosedInterval));
...
```

jcphill: I'm trying to avoid clearing the same interval twice (I can't find anything in the standard that says that is legal). I've also been thinking about closures in C++ instead of JavaScript. Modified to also remove the dispose handler.

src/navigators/PopupWindow.ts
jcphill: Fixed.

src/navigators/PopupWindow.ts
pamapa: Ah i understand. I have also not found something in the standard. Your code looks good enough. Multiple calls to clearInterval with same id seems to work fine...

pamapa:
tests fail with:
```
 A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

Summary of all failing tests
FAIL src/navigators/PopupWindow.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of null (reading '_history')

      at Window.get history [as history] (node_modules/jsdom/lib/jsdom/browser/Window.js:379:79)
          at Array.forEach (<anonymous>)
```

pamapa:
I have tested with sample/Parcel the "with parent" path and explicitly set parent to `null` in `_notifyParent` to test "no parent" and both worked.

jcphill:
> ```
>     TypeError: Cannot read properties of null (reading '_history')
> 
>       at Window.get history [as history] (node_modules/jsdom/lib/jsdom/browser/Window.js:379:79)
>           at Array.forEach (<anonymous>)
> ```
Sorry, I don't have a way to debug this as the tests all run fine in my build environment but I've attempted a fixed based on the error message and similarity to how window.location is handled.

pamapa:
now it fails in the github pipeline with this:
```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
Summary of all failing tests
FAIL src/navigators/PopupWindow.test.ts
  ● Test suite failed to run
    TypeError: Cannot read properties of null (reading '_origin')
      at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:424:50)
          at Array.forEach (<anonymous>)
Test Suites: 1 failed, 31 passed, 32 total
```

jcphill:
Try it again, please.

jcphill:
I managed to reproduce the CI issues by switching to node 18 and it should work now.

I also fixed up some tests that were leaving BroadcastChannel handles open. Two of them were described as "should never resolve" so no surprise, but "and data url parse fails should reject with error" absolutely looked like it should be calling this._dispose() before rejecting. The problem was that the mock for window.addEventListener called the listener immediately so this._dispose() was called before the channel close handler was added.

src/navigators/AbstractChildWindow.ts
```diff
@@ -87,11 +90,24 @@ export abstract class AbstractChildWindow implements IWindow {
...
             source: messageSource,
             url,
             keepOpen,
-        } as MessageData, targetOrigin);
+        };
+        if (parent) {
+            parent.postMessage(msgData, targetOrigin);
+        } else {
+            const logger = new Logger("_notifyParent");
+            logger.debug("No parent. Using BroadcastChannel.");
```
pamapa, line 103: When i was debugging this i immediately added a log in the `if (parent)` code path as well. Please do so too.
Something like:
```
...
const logger = new Logger("_notifyParent");
if (parent) {
    logger.debug("With Parent. Using parent.postMessage.");
} ...
```

jcphill: Done.

pamapa (approved):

pamapa:
Thank you for your patience and your contribution!

jcphill:
> Thank you for your patience and your contribution!

The same to you!

nattb8:
Thanks for the fix @jcphill ! 

@pamapa  - do you have an ETA for when this fix might be released? It's quite urgent for us to get this in a proper release. Thanks!

pamapa:
> Thanks for the fix @jcphill ! I tested it by pointing our project to the latest `main` branch and can confirm it resolves the issue on our end.
> 
> @pamapa - do you have an ETA for when this fix might be released? It's quite urgent for us to get this in a proper release. Thanks!

@nattb8 and @jcphill: I made a release candidate (v3.3.0-rc.0), which will become the release after testing. Please give feedback...

jcphill:
The release candidate works for me.

PR #1925, opened 2025-04-04
fix: Resource Owner Password Credentials flow with oidc scope and no id_token returned will fail sub claim check

sularome:
If we have a Resource Owner Password Flow and token endpoint does not return `token_id` in the response while returning scope: `openid` it will fail the sub claim check.
```
        if (validateSub && claims.sub !== response.profile.sub) {
            logger.throw(new Error("subject from UserInfo response does not match subject in ID Token"));
        }
```
After some investigation, i noticed that if we don't have token_id ``_validateIdTokenAttributes`` will be skipped which means that ``response.profile`` will always be an empty object and always fail the check ``(validateSub && claims.sub !== response.profile.sub)``

### DIFF
--- a/src/ResponseValidator.test.ts
+++ b/src/ResponseValidator.test.ts
@@ -610,6 +610,22 @@ describe("ResponseValidator", () => {
 
         });
 
+        it("should process an openid signin response without an id_token and userinfo containing sub property, as a non-openid signin response", async () => {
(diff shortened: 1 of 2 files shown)


pamapa (approved):

pamapa:
thanks for contributing

PR #2307, opened 2025-12-22
feat: migrate test framework from jest to vitest

SanseroGames:
Jest has some issues when working with ES-modules and needed some workarounds for working with jsdom. ViTest is a more modern drop-in replacement for Jest that solves those issues. For example it is no longer necessary to configure a custom JSDom environment, as all required properties are now exposed with the new setup.

This PR does not close any issues. The change has been discussed with @pamapa 

### Checklist


pamapa (approved):

pamapa:
Thanks for contributing!

PR #2460, opened 2026-05-22
Support custom iframe attributes for silent signin

Sune1337:
This PR adds support for passing custom HTML attributes to the iframe used during silent signin (signinSilent).

**Motivation**
Recent Chrome changes around Local Network Access (PNA / Private Network Access) require explicit permission delegation for iframe-based requests to local or private network resources.

A common real-world scenario is:

> A web application hosted on a private/internal address (e.g. https://app.local or an internal DNS/IP) using a public identity provider such as Microsoft Entra ID (Azure AD).

In this setup:

- The main application runs on a private/local network
- Authentication happens via a public endpoint (e.g. login.microsoftonline.com)
- Silent signin is performed using a hidden iframe

With the latest browser restrictions, this iframe may be blocked unless it explicitly declares permission via:
```html
<iframe allow="local-network-access *">
```

Without this attribute:
- Silent authentication can fail
- Users may be unexpectedly logged out or forced into full redirects

**Changes**
Adds optional iframeAttributes configuration:
- Available globally via UserManager settings
- Available per call via signinSilent arguments

**Example usage**
Global configuration
```ts
const userManager = new UserManager({  iframeAttributes: {    allow: "local-network-access *"  }});
```

Per-call override
```ts
await userManager.signinSilent({  iframeAttributes: {    allow: "local-network-access *"  }});
```

**Notes**

- Using the feature is optional and should not change any behavior if not used.
- Supports arbitrary attributes, allowing future browser requirements to be handled without further changes

### DIFF
--- a/docs/oidc-client-ts.api.md
+++ b/docs/oidc-client-ts.api.md
@@ -132,6 +132,8 @@ export interface IdTokenClaims extends Mandatory<OidcStandardClaims, "sub">, Man
 // @public (undocumented)
 export interface IFrameWindowParams {
     // (undocumented)
+    iframeAttributes?: Record<string, string>;
+    // (undocumented)
     silentRequestTimeoutInSeconds?: number;
 }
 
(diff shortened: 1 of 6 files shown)


pamapa (approved):

pamapa:
looks good, thanks for contributing

PR #2464, opened 2026-06-12
Update proxy target URL for OIDC

boborua:
Closes/fixes #issue

### Checklist

### DIFF
--- a/samples/Parcel/.proxyrc.js
+++ b/samples/Parcel/.proxyrc.js
@@ -4,7 +4,7 @@ module.exports = function (app) {
     app.use(
         "/oidc/",
         createProxyMiddleware({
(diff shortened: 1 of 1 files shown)


pamapa:
why is this needed?

boborua:
> why is this needed?

Behavior: 
without this modification: 
image

This is cause by different behavior of `http-proxy-middleware` v2 and v3. [See this doc](https://github.com/chimurai/http-proxy-middleware/discussions/768)

As v3 is specified in parent project's package.json, so this place should be modifed.
Maybe in sample/Parcel sub project, `http-proxy-middleware@v3` should also be declared?

pamapa (approved):

pamapa:
thanks for contributing!

PR #1486, opened 2024-04-27
feat(User): Add flexibility to User class.

benMain:
Sometimes more data needs to be captured from the Token endpoint response than is standard. I'm proposing allowing users to specify extraTokenResponseKeys which will populate an extraTokenResponseProperties object inside the User object. 

Closes/fixes #1485
https://github.com/authts/oidc-client-ts/issues/1485

### Checklist

- [x] This PR makes changes to the public API 
- [x] I have included links for closing relevant issue numbers

PR #2471, opened 2026-06-24
feat(extraTokenProperties): Add extra token properties

### LINKED ISSUE #1485: How to get additional AccessToken Response properties from SigninResponse to User object, or make user object extensible.
I'm dealing with an OIDC situation where the access token response contains more useful properties that I need than get exposed by the User object. [EPIC OAUTH implementation](https://fhir.epic.com/Documentation?docId=oauth2&section=Embedded-Oauth2-Launch_Access-Token-Request_No_Refresh-Tokens). For example their token response includes a patient property that would be very useful if it could be stored in the User object. UserManagerSettings seems to allow for a lot of overrides, but not which properties should be retained from the SigninResponse for the User object.

I know that they exist in the SigninResponse object, but the constructor of the user object filters that to the known properties. I like working with the events that come off the UserManager, it makes working your library easy. If I'm missing something please tell me, but I feel like we should have a setting to extend the User object. 

I would be happy to work on that if interested.

pamapa:
Takes code changes from merge-request #1486 and leaves not wanted changes like packge.json, etc... away, and improves code changes according to review in this merge-request.

Closes/fixes #1485

### Checklist

### DIFF
--- a/src/User.ts
+++ b/src/User.ts
@@ -49,6 +49,12 @@ export class User {
     /** The expires at returned from the OIDC provider. */
     public expires_at?: number;
 
+    /**
+     * The additional token response properties the user has requested to keep via settings.
+     * @see {@link UserManagerSettings.extraTokenResponseKeys}
+     */
+    public extraTokenResponseProperties?: Record<string, unknown>;
+
     /** custom state data set during the initial signin request */
     public readonly state: unknown;
     public readonly url_state?: string;
@@ -64,7 +70,7 @@ export class User {
         expires_at?: number;
         userState?: unknown;
         url_state?: string;
-    }) {
+    }, extraTokenResponseKeys?: string[]) {
         this.id_token = args.id_token;
         this.session_state = args.session_state ?? null;
         this.access_token = args.access_token;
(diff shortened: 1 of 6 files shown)


pamapa:
@Laureatgrepi Time to verify/review this changes?

src/UserManager.ts
```diff
@@ -548,7 +548,7 @@ export class UserManager {
 
     protected async _buildUser(signinResponse: SigninResponse, verifySub?: string) {
         const logger = this._logger.create("_buildUser");
-        const user = new User(signinResponse);
+        const user = new User(signinResponse, this.settings.extraTokenResponseKeys);
```
Laureatgrepi, line 551: Do you need to pass this argument inside the `_useRefreshToken()` as well?

src/User.ts
```diff
@@ -76,6 +82,16 @@ export class User {
         this.expires_at = args.expires_at;
         this.state = args.userState;
         this.url_state = args.url_state;
+
+        if (extraTokenResponseKeys?.length) {
+            this.extraTokenResponseProperties = {};
+            for (const key of extraTokenResponseKeys) {
+                const extraToken = (args as Record<string, unknown>)[key];
+                if (extraToken) {
```
Laureatgrepi: Would maybe variable something else than `extraToken`. Maybe `extraProperty`?

Also what happens here if the actual value of the property is `false` or `0` or something similar? This would need to be taken in to consideration.

Laureatgrepi (approved):
Didn't verify that this works but LGTM

PR #2474, opened 2026-08-04
fix: export RefreshState class instead of only its type

### LINKED ISSUE #2433: RefreshState not exported from ESM/UMD bundles despite being part of the public API
RefreshState is marked @public in the source and is included in the type declarations (export declare class `RefreshState`).
It's also required to call the public OidcClient.useRefreshToken() method, since `UseRefreshTokenArgs.state` is typed as `RefreshState`.

However, `RefreshState` is not actually exported from the ESM, UMD, or browser bundles. 
This means consumers using `OidcClient` directly (rather than UserManager) cannot construct a RefreshState at runtime to call `useRefreshToken`.

  **Steps to reproduce:**

  1. Import RefreshState from oidc-client-ts in a project using the ESM bundle (e.g. via Vite/Rollup)
  2. Build fails because RefreshState is not exported from the bundle

  import { OidcClient, RefreshState } from 'oidc-client-ts';
  // RefreshState is undefined at runtime

** Repro repo**
[https://github.com/eugenioenko/oidc-client-ts-refresh-state-bug-repro](https://github.com/eugenioenko/oidc-client-ts-refresh-state-bug-repro)

Stackblitz [https://stackblitz.com/github/eugenioenko/oidc-client-ts-refresh-state-bug-repro](https://stackblitz.com/github/eugenioenko/oidc-client-ts-refresh-state-bug-repro), just run `npm run build`

```
src/main.js (1:21): "RefreshState" is not exported by "node_modules/oidc-client-ts/dist/esm/oidc-client-ts.js", imported by "src/main.js".
```

  **Expected behavior:**

  RefreshState should be exported from all bundles, since it's required to use the public useRefreshToken API on OidcClient.

  **Workaround**:

  Construct a plain object matching the `RefreshState` shape and cast it via UseRefreshTokenArgs['state']. 

  Affected versions: 3.2.1, 3.5.0 (latest)

eugenioenko:
Fixes #2433

```ts
export type { RefreshState } from "./RefreshState";
```

`RefreshState` was exported with `export type`, which gets erased at compile time. This made the class `undefined` in the ESM/UMD/browser bundles even though it's a public class that consumers need to instantiate to call `OidcClient.useRefreshToken()`.

The fix changes it to a regular export so the runtime value is preserved:

```ts
export { RefreshState } from "./RefreshState";
```

Added `src/index.test.ts` to verify that `RefreshState` is exported as a runtime value and can be constructed when imported from the public entry point.


pamapa:
not able to merge: the unit-test is failing:

```
src/index.test.ts:20:17 - error TS2739: Type '{ sub: string; }' is missing the following properties from type 'IdTokenClaims': iss, aud, exp, iat

20                 profile: {
                   ~~~~~~~

  src/RefreshState.ts:26:9
    26         profile: UserProfile;
               ~~~~~~~
    The expected type comes from property 'profile' which is declared here on type '{ refresh_token: string; id_token?: string | undefined; session_state: string | null; scope?: string | undefined; profile: IdTokenClaims; state?: unknown; }'
```

It would be ok when you send the fix without a unit-test in that particular case...

eugenioenko:
> It would be ok when you send the fix without a unit-test in that particular case...

@pamapa , I've update the PR to remove the extra test. Agreed.

pamapa (approved):

pamapa:
thanks for contributing

